PR #2, opened 2015-06-03
Add preliminary scaffolding for server routes and logging

gideonthomas:


app.js
```diff
@@ -0,0 +1,50 @@
+/* This creates a Hapi server which provides a RESTful API to allow Thimble to
```
sedge: Not just thimble - goggles too.

README.md
```diff
@@ -1,2 +1,37 @@
 # publish.teach.org
 The teach.org publishing service for goggles and thimble
```
humphd: "X-Ray Goggles and Thimble"

app.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+}, function(err) {
+  if(err) {
+    log.disabled = true;
+    console.warn('Failed to load logging plugin, disabling logging');
+  }
+
+  server.start(function(err) {
+    if(err) {
+      log.fatal('Failed to start the server');
```
humphd: Can you also log the `err` object here so we know what happened?

routes/index.js
```diff
@@ -0,0 +1,20 @@
...
+  // Handles user-level operations
+  require('./projects/get-projects.js')(server);
+  require('./projects/create-project.js')(server);
+  require('./projects/delete-project.js')(server);
+  require('./projects/publish-project.js')(server);
+
+  // /files/* API routes
+  // Handles project-level operations
+  require('./files/get-project-contents.js')(server);
+  require('./files/get-file-contents.js')(server);
```
humphd: I wonder if you should split out `project/` and `file/` so you can get project contents or file contents?  So you'd have routes for:
* `/projects/...`
* `/project/...`
* `/file/...`

gideonthomas: i.e.:
* `projects/:username` - returns a list of projects
* `project/:username/:projectName/*` - crud and publish **a** project
* `file/:username/:projectName/*` - crud on files in a project

Is that what you have in mind?

humphd: Exactly, what do you think?  It seems like you have those 3 high level things to worry about, might as well make them all separate routes.

gideonthomas: Agree! That's probably cleaner

sedge: +1

authentication/index.js
```diff
@@ -0,0 +1,2 @@
+module.exports = function(server) {
+};
```
humphd: Can you throw a comment in here at least as to what this will be or do?

routes/index.js
```diff
@@ -0,0 +1,23 @@
...
+  // /projects API route
+  // Handles user-level operations
+  require('./projects/get-projects.js')(server);
+
+  // /project/* API routes
+  // Handles project-level operations
+  require('./project/get-contents.js')(server);
+  require('./project/create.js')(server);
+  require('./project/delete.js')(server);
+  require('./project/publish.js')(server);
```
humphd: Do we need a `rename` here?  We probably don't want to `delete` and `create` if that means you lose data.

routes/index.js
```diff
@@ -0,0 +1,23 @@
...
+
+  // /file/* API routes
+  // Handles project-specific file-level operations
+  require('./file/get-contents.js')(server);
+  require('./file/create-update.js')(server);
+  require('./file/delete.js')(server);
```
humphd: Same question here, do we want `rename`?

routes/project/publish.js
```diff
@@ -0,0 +1,12 @@
...
+ * Publishes the specified project
+*/
+
+module.exports = function(server) {
+  server.route({
+    method: 'POST',
+    path: '/project/{username}/{projectName}/publish',
+    handler: function(request, reply) {}
+  });
+};
```
humphd: What about `unpublish`?

routes/index.js
gideonthomas: How would you lose data?

humphd:
I've never used Hapi, so best to get some follow-up review from @cadecairos and/or @jbuck when you think it's ready.

routes/project/publish.js
gideonthomas: Are we implementing that? I thought we nixed it!

routes/index.js
humphd: If `delete` actually deletes a project, then you've lost your data when you do `delete` + `create`, where data is any linked content in other tables (e.g., files).  It would force you to re-upload.

authentication/index.js
cadecairos: For authentication, you'll probably want to go the scheme/strategy [1] route,  and load it as a plugin [2] rather than pass your server object into an exported function

1. http://www.hapijs.com/tutorials/auth
2. http://www.hapijs.com/tutorials/plugins

app.js
```diff
@@ -0,0 +1,43 @@
...
+var log = require('./logger.js');
+
+var server = new Server();
+server.connection({ port: env.get('port') });
+
+// Configure authentication middleware
+authentication(server);
+
+// Configure server end-points
+routes(server);
```
cadecairos: Similar thing here,  I recommend wrapping your routes in a registered plugin for the server

routes/index.js
gideonthomas: renames will be part of `create-update`

gideonthomas:
@cadecairos is this what you had in mind?

routes/project/publish.js
gideonthomas: added unpublish

cadecairos:
@gideonthomas :+1: yup, that's a hapi server right there.

sedge:
r+

Issue #3, opened 2015-06-05
Finalize publish.wm.o API


sedge:
# API for thimble to create, replace, update, and delete projects (publish.webmaker.org)

## Schemas
```js
project: 
{
  name: ...,
  username: ...,
  isIndexable: bool,
  files: [
    {...}
  ],
  publishedAt: DateTime
}

file:
{
  type: "content-type",
  path: "/user/project/nested/index.html",
  link: "https://link/to/file.jpg"
}
```

## RESTful API
GET /projects/:username
* List the given users projects
```javascript
{
  projects: [
    ...
  ]
}
```

GET /projects/:username/:projectname
* Retrieve a single project object

```javascript
{
  projects: ...
}
```
POST /projects
* Create a project for the given user and project not including 
```js
{ Project Object }
```

DELETE /projects/:username/:projectname
* Delete a project for the given user and project

PUT /projects/:username/:projectname
* Updates metadata about a project, including publish/unpublish/name

GET /projects/:username/:projectname/files/:filename
* Get the content for the given file

PUT /projects/:username/:projectname/files/:filename
* Create or replace the content for the given file, with the given Content-Type

DELETE /projects/:username/:projectname/files/:filename
* Delete the given file

## Remote procedure call routes
PUT /projects/:username/:projectname/publish
* Publishes a project by copying the saved version to a publicly accessible end-point. If the project is already published, republish with the working copy.

PUT /projects/:username/:projectname/unpublish
* Unpublishes a project

PUT /projects/:username/:projectname/rename
* Renames a project
```js
{
  newName: ...
}
```

PUT /projects/:username/:projectname/files/:filename/rename
* Renames a file
```js
{
  newName: ...
}
```

humphd:
cc @flukeout, @simonwex 

I think the way this imagines storing a project wrt it being saved-only vs. saved-and-published is wrong.  Luke wants to see us use a publish-creates-a-snapshot-copy vs. publish-flips-an-access-pref.  The reason for this is so that a user can publish1, save1, save2, save3, then publish again, and if save1 and save2 broke the code, it won't affect the published content.  In other words, publish needs to take whatever is in the project (all files) and *copy* them to some new place from which they'll be served.  Subsequent saves won't affect the published content at all.

If I'm misreading the above in regard to this, apologies.  If I'm not misreading, we need to update.

sedge:
[...]

PR #6, opened 2015-06-07
Update API routes and validation schemas

gideonthomas:
Updated routes based on [this discussion](https://github.com/mozilla/publish.webmaker.org/issues/3)


routes/projects/files/create-update.js
```diff
@@ -0,0 +1,27 @@
+/*
+ * PUT /projects/:username/:projectName/files/:pathToFile
+ * The :pathToFile parameter must be URI encoded
+ * Create or replace the contents for the given file path
+ * The message body that would normally be expected by this route
+ * will look something like:
+ * {
+ *   type: <string> content-type,
+ *   path: <string>,
+ *   data: <string> base64 encoded binary data
```
humphd: It would be nice if this could be `String` (utf8) or `ArrayBuffer` (binary).  Having to deal in base64 for binary data sucks.

routes/projects/files/rename.js
```diff
@@ -0,0 +1,23 @@
...
+ *   name: <string> required
+ * }
+*/
+
+var renameSchema = require('../../../schemas.js').rename;
+
+module.exports = function(server) {
+  server.route({
+    method: 'PUT',
+    path: '/projects/{username}/{projectName}/files/{path}/rename',
```
humphd: Can one rename a file, and therefore `mv` it to a new path?  Or is rename always a change of name only for the leaf node?

routes/projects/files/get-one.js
```diff
@@ -0,0 +1,18 @@
+/*
+ * GET /projects/:username/:projectName/files/:pathToFile
+ * The :pathToFile parameter must be URI encoded
+ * Returns the contents of the given file in the form:
+ * {
+ *   type: <string> content-type,
+ *   path: <string>,
+ *   link: <string>
+ * }
```
humphd: I wonder if it's useful to also send `size` so that clients can judge the amount of work to download various resources (e.g., progress updates for opening a project).

routes/projects/update.js
```diff
@@ -0,0 +1,28 @@
...
+ * PUT /projects/:username/:projectName
+ * Creates a new project for the given user. A project object is sent in
+ * the form:
+ * {
+ *   name: <string> required,
+ *   username: <string> required,
+ *   isPublic: <boolean> required,
+ *   publishedAt: <dataTime> optional,
+ *   publishLink: <string> optional,
+ *   files: [ <FileObj1>, <FileObj2>... ] required
```
humphd: Can this list of files be partial?

ashleygwilliams:
looks like this would fix #8

routes/projects/files/rename.js
sedge: Rename is always a change of name, not the full path.

routes/projects/files/get-one.js
sedge: +1

routes/projects/files/rename.js
ashleygwilliams: this is an interesting question, but i think we need to remember that a path and the name of a file are actually the same, e.g:

`/public/styles/custom.css` is the true name of `custom.css`

routes/projects/update.js
sedge: Yes - further files should be added through `PUT /projects/:username/:projectname`

sedge: Though I wonder if this should be optional?

routes/projects/files/rename.js
humphd: So we'll have to do some work to maintain consistency across the backend storage view of a file vs. the filesystem, where the filesystem will only ever let us rename the leaf node.  It might be useful for us to disallow renaming `/public/styles/custom.css` to `/public/custom.css`, since it will lead to filesystem data loss.

routes/projects/files/create-update.js
gideonthomas: For 'string' data, do you mean actual strings and not binary (encoded) data?

humphd: I'd suggest we always send/receive `ArrayBuffer` on the browser side, and always send/receive `Buffer` on the server side.  We don't need to encode at all.

gideonthomas: Actually, it's just my documentation that is wrong haha

routes/projects/update.js
gideonthomas: It should basically represent whatever is in the user's fs (on the thimble side) at the moment this request is sent. So values can range between `[ ]` to `[ file1, file2, file3... file100 ]`

sedge: Even if those files aren't uploaded to the server? Doesn't a file object need to point at the file data?

gideonthomas: Hmm this does seem a bit problematic. @ashleygwilliams do you have any insight into this? What kind of foreign key validation do we want to do?

sedge: Since a foreign key won't exist yet, my vote is to keep it simple - Creating a project means it has no files associated with it yet. Adding files through the `file` routes can update the related `project` object.

gideonthomas: This is a PUT route though, which means updating project metadata will need you to provide all the list of files associated with the project...which can be problematic

routes/index.js
```diff
@@ -4,25 +4,23 @@
...
+  require('./projects/files/delete.js')(server);
 
-  // /file/* API routes
-  // Handles project-specific file-level operations
-  require('./file/get-contents.js')(server);
-  require('./file/create-update.js')(server);
-  require('./file/delete.js')(server);
-  require('./file/rename.js')(server);
+  // RPC routes
+  require('./projects/rename.js')(server);
```
humphd: Question: why do you separate the `*rename` routes out like this, rather than including with the associated routes above?

routes/projects/update.js
sedge: If we consider "files" only to exist if they exist on the server then this should be okay.

routes/index.js
sedge: This was because they aren't REST endpoints, since they're fire and forget operations. They're meant to condense the process of renaming down to one call.

gideonthomas: Hmm I guess I thought it fit well with the type of route it is. But it probably makes more sense to associate it with the data it is working on i.e. project or file

gideonthomas: Well it honestly is a matter of preference, so either way works

routes/projects/update.js
gideonthomas: well that means that the client needs to make sure that we only send files that are on the server and not "queued up for autosave"

routes/projects/create.js
```diff
@@ -0,0 +1,28 @@
+/*
+ * POST /projects
+ * Creates a new project for the given user. A project object is sent in
+ * the form:
+ * {
+ *   name: <string> required,
+ *   username: <string> required,
+ *   isPublic: <boolean> required,
+ *   publishedAt: <dataTime> optional,
+ *   publishLink: <string> optional,
```
humphd: The client won't send `publishedAt` or `publishLink`, so I think this is confusing.  It will get *stored* with this data, but I don't think you want to require it to be *sent* in this form.

I think there are optional bits to this.  For example:

```js
{
  name: "...",
  username: "...",
  files: [initial, list, of, files]
}
```

In this case, `isPublic` can take the value `false` since it is absent, and `publishedAt` and `publishLink` will get updated later.

routes/projects/update.js
sedge: How so? AFAIK it just means that we populate the project object with file references as we upload them to the server for the first time.

routes/projects/delete.js
```diff
@@ -1,12 +1,12 @@
 /*
- * DELETE /project/:username/:projectName
+ * DELETE /projects/:username/:projectName
  * Deletes the given project for the given user
```
humphd: I assume this cascades to also delete all files associated with the project, as well as published resources?  If so, call it out here.

routes/projects/files/create-update.js
```diff
@@ -0,0 +1,28 @@
+/*
+ * PUT /projects/:username/:projectName/files/:pathToFile
+ * The :pathToFile parameter must be URI encoded
+ * Create or replace the contents for the given file path
+ * The message body that would normally be expected by this route
+ * will look something like:
+ * {
+ *   type: <string> content-type required,
+ *   path: <string> required,
+ *   size: <integer> required,
```
humphd: This isn't necessary, since it can be determined by the size of the body contents.  I'd remove it.

routes/projects/update.js
sedge: A sequence diagram would help here -.-

1. Create a project with POST /projects
2. Upload existing files for that with PUT /projects/:username/:projectname/files/:filename, which adds them as foreign keys in the project object

ashleygwilliams: ok- so a few thoughts:

when a project is created, i assume that a user will be redirected to a view that has a file open and ready to edit (likely `index.html`). this default file should be created at the time the project is created. that way a foreign key will exist. all projects, to be projects need at least one file, IMO.

thoughts?

routes/projects/files/get-one.js
```diff
@@ -0,0 +1,19 @@
+/*
+ * GET /projects/:username/:projectName/files/:pathToFile
+ * The :pathToFile parameter must be URI encoded
+ * Returns the contents of the given file in the form:
+ * {
+ *   type: <string> content-type,
+ *   path: <string>,
+ *   size: <integer>,
+ *   link: <string>
```
humphd: What is `link`?  Where is the actual `Buffer` for this file?

routes/projects/files/create-update.js
gideonthomas: as in make it optional?

routes/projects/files/rename.js
humphd: @gideonthomas I think you need to update the comment above to indicate that we only allow `filename` changes, not `pathname`.

routes/projects/files/get-one.js
gideonthomas: I should probably add a little more documentation here. I think our plan was to provide a `link` to the file contents (maybe in s3) as opposed to adding it to the metadata
cc @ashleygwilliams

routes/projects/update.js
ashleygwilliams: re: updating files- new files should be created at `POST /files`, not at `PUT /projectname`. though we could implement a `PATCH /projectname` but that gets gross since the project should really only contain the foreign keys, not the actual file object representation...

tl;dr files should be made and updated at `POST/PUT /files`. they'll be associated with the project by foreign key. so when the `files` attribute is passed in a `PUT /projectname` it shouldn't be updating anything.

humphd: I think this API's view of files associated with a project has to relate to what's on the server, which is to say, I don't think a client should be able to specify it, only create, delete, and rename files for a project.  Since this payload won't include those files, I'd suggest you remove this, and just use `[]` until the files themselves are uploaded.

sedge: @ashleygwilliams Hmm. Projects will be populated by default, but only on the client side since they won't exist as "project objects" until `POST /projects` is hit. At that point, the files haven't been uploaded to the server yet, so I don't see how there can be reference to them.

Unless we want to create a new "project object" every time a user opens thimble. With a generic name maybe?

sedge: @humphd +1

routes/projects/files/get-one.js
ashleygwilliams: yes that is correct. i think serving a blob from the API is not a great idea. passing a representation of the file and not the file itself seems best in my opinion. the blob should be served statically.

sedge: It looks like we're leaning towards postgres. I think it would be simplest to just return the buffer here.

routes/projects/update.js
gideonthomas: So what's the verdict on this?
POST /projects body:
```js
{
  ...
  files: []
}
```
PUT /projects/:username/:prjName
```
same as above
```
GET /projects/:username/:prjName
```
{
  ...
  files: [ 
            { 
               ..., 
               link: linkTofile 
            }, 
            ... 
         ]
}
```

Is that it?
@ashleygwilliams @humphd @sedge

sedge: We need to figure out how we're going to access the data for a file before we can definitively answer this question.

routes/projects/delete.js
sedge: Yeah it would.

schemas.js
```diff
@@ -0,0 +1,25 @@
+/*
+ * Validation schemas for incoming requests
+*/
+
+var Joi = require('joi');
+var mimes = Object.keys(require('mime-db'));
+
+exports.file = Joi.object().keys({
+  type: Joi.string().valid(mimes),
```
sedge: Don't we have restrictions on which files we'll accept into a project? This would be a good place to enforce that.

gideonthomas: Are there restrictions on which files we'll accept? Anyway, this is not going to be taken in the PUT/POST so the client doesn't really have control over this, the server determines what the mime type is.

ashleygwilliams: IMO that would be an enhancement. let's not worry about that for now, ship a simple version of this and move the thread to the issues.

sedge: +1

Issue #23, opened 2015-06-08
Scaffold the DB + schemas


sedge:
Our DB layer needs to be put together for use with the API.

ashleygwilliams:
this was completed in #27, closing.

PR #24, opened 2015-06-08
[WIP] Data layer

ashleygwilliams:
work in progress, please do not merge.

solves #23

### DIFF
--- a/migrations/1_create_users_projects_files.js
+++ b/migrations/1_create_users_projects_files.js
@@ -1,0 +1,27 @@
+"use strict";
+
+module.exports = {
+  up: function(migration, DataTypes, done) {
+    migration.createTable(
+      'Users', {
+        name: DataTypes.STRING
+      });
+    migration.createTable(
+      'Projects', {
+        title: DataTypes.STRING,
+        user_name: DataTypes.STRING
+      });
+    migration.createTable(
+      'Files', {
+        path: DataTypes.STRING,
+        type: DataTypes.STRING,
+        project_title: DataTypes.STRING
+      })
+  },
+
+  down: function(migration, DataTypes, done) {
+    migration.dropTable('Users');
+    migration.dropTable('Projects');
+    migration.dropTable('Files');
+  }
+};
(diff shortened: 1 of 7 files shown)


migrations/1_create_users_projects_files.js
```diff
@@ -0,0 +1,27 @@
...
+module.exports = {
+  up: function(migration, DataTypes, done) {
+    migration.createTable(
+      'Users', {
+        name: DataTypes.STRING
+      });
+    migration.createTable(
+      'Projects', {
+        title: DataTypes.STRING,
+        user_name: DataTypes.STRING
```
gideonthomas: `is_public: DataTypes.BOOLEAN

migrations/1_create_users_projects_files.js
```diff
@@ -0,0 +1,27 @@
...
+      });
+    migration.createTable(
+      'Files', {
+        path: DataTypes.STRING,
+        type: DataTypes.STRING,
```
gideonthomas: I don't know if we want to store the content type in the db. I think we should determine it based on the file extension on-the-fly when we serve it

models/file.js
```diff
@@ -0,0 +1,16 @@
+"use strict";
+
+module.exports = function(sequelize, DataTypes) {
+  var Task = sequelize.define("File", {
```
gideonthomas: `File` or `Files`? I think we should use singular for this

models/index.js
```diff
@@ -0,0 +1,31 @@
+"use strict";
+
+const fs        = require("fs");
+const path      = require("path");
+const Sequelize = require("sequelize");
+const env       = process.env.NODE_ENV || "development";
```
gideonthomas: `require('./environment.js').get('node_env');` since we have an `environment.js` config file. If you want it to be set from command line, probably add logic to that file similar to `LOG_LEVEL`

ashleygwilliams:
closing in favor of using bookshelf/knex as per convo with @cadecairos and agreed upon by the rest of the team

PR #31, opened 2015-06-09
add bookshelf models and GET element and collection routes

ashleygwilliams:


ashleygwilliams:
review plz :bow:

server.js
```diff
@@ -0,0 +1,29 @@
+require('habitat').load();
```
gideonthomas: This is already done in `environment.js`

server.js
```diff
@@ -0,0 +1,29 @@
...
+  if ( err ) {
+    server.log('error', {
+      message: 'Error registering API',
+      error: err
+    });
+    throw err;
+  }
+
+  server.start(function() {
+    console.log('Server started @ ' + server.info.uri);
```
gideonthomas: Prefer bunyan logging:
```
var log = require('./logger.js');
log.info({server: server.info}, 'Server started');

Comment (gideonthomas) [inline server.js]: And you might want to register the logger as a plugin as well so that every log passes through bunyan

File: server.js [hunk 3]
```diff
@@ -0,0 +1,21 @@
...
+
+var connection = {
+  host: process.env.API_HOST,
+  port: process.env.PORT
+};
+
+const server = new Hapi.Server();
+server.connection(connection); 
+
+server.register(require('./api'), function(err) {
```
gideonthomas: you might also want to register the logger as a plugin so that we capture server logged events:
```javascript
register: hapiBunyan,
options: {
      logger: log
}
```

gideonthomas:
r=me

api/classes/database.js
```diff
@@ -0,0 +1,16 @@
+const config = require('../../knexfile')['development'];
+
+// reset database at each startup
+//try {
```
sedge: Why is this commented out? I'd rather see it behind an `env` check

api/classes/database.js
```diff
@@ -0,0 +1,16 @@
...
+//  require('fs').unlinkSync(config.connection.filename);
+//} catch (e) {}
+
+const Knex = require('knex')(config);
+const Bookshelf = require('bookshelf')(Knex);
+
+//console.log('Running migrations...');
+//Knex.migrate.latest(config).then(function () {
```
sedge: ditto

api/classes/log.config
```diff
@@ -0,0 +1,10 @@
+var bunyan = require('bunyan');
```
sedge: What is this file?

api/modules/files/model.js
```diff
@@ -0,0 +1,22 @@
+const BaseModel = require('../../classes/base_model');
+
+const instanceProps = {
+  tableName: 'files',
+  comic: function () {
```
sedge: `comic`?

api/modules/files/model.js
```diff
@@ -0,0 +1,22 @@
...
+    project_id: function (qb, value) {
+      return qb.whereIn('project_id', value);
+    }
+  },
+  relations: [
+    'project'
+  ]
+};
+
+module.exports = BaseModel.extend(instanceProps, classProps);
```
sedge: Now that I've taken a closer look at this, I really like the way these relations are declared. Is this also what defines them in the db itself?

api/classes/database.js
ashleygwilliams: for dev purposes you might want to drop the db and seed/migrate, if you uncomment this, it will do that on npm start

env.dist
```diff
@@ -2,11 +2,13 @@
 # Config for publish.webmaker.org server
 ##
 
-# development or production
+# Development or Production
 export NODE_ENV="development"
+export API_VERSION=dev
```
sedge: Where do we use this?

api/modules/files/model.js
ashleygwilliams: lol that's from jocular, i'll fix it

api/modules/files/model.js
ashleygwilliams: no, the db stuff is defined in the migration file.

api/classes/database.js
sedge: I've been taught to hate commented code with violent prejudice but I can live with this ;)

ashleygwilliams: frankly, i think i should just delete this, it's from a demo i did for endpoints and doesn't really apply here- agree?

sedge:
Looking really good. This architecture makes a lot of sense to me. Nice one!

api/classes/database.js
sedge: If you think it doesn't belong, nuke it.

sedge:
r+

ashleygwilliams:
thanks ya'll! :whale:

ashleygwilliams:
lol :disappointed: i forgot to squash

PR #32, opened 2015-06-09
Fix 22 - Implement test scaffolding

sedge:
This is a work in progress. The main point of contention will be the fixtures and their relationship to our seed scripts.

Remaining work:

* [x] NPM scripts (populate db with test data and run tests at a command)

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "start": "node server",
     "env": "cp env.dist .env",
     "db": "knex migrate:latest && knex seed:run",
-    "test": "lab",
+    "test": "rm dev.db && npm run db && lab -t 100 -c --verbose --colors --assert code --timeout 5000 && npm run lint",
     "lint": "npm run jshint && npm run jscs",
     "jshint": "jshint -c node_modules/mofo-style/linters/.jshintrc api authentication migrations seeds logger.js server.js",
     "jscs": "jscs -c node_modules/mofo-style/linters/.jscsrc api authentication migrations seeds logger.js server.js"
@@ -41,6 +41,8 @@
   },
   "devDependencies": {
     "bunyan-prettystream": "^0.1.3",
+    "code": "^1.4.0",
+    "cross-spawn": "^0.4.0",
     "jscs": "^1.13.1",
     "jshint": "^2.8.0",
     "lab": "^5.9.0",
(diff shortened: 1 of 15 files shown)


sedge:
If this is horribly over-engineered I'll simplify it. I tried to copy `api.wm.o`'s approach as much as possible.

test/lib/fixtures/files/get-one.js
```diff
@@ -0,0 +1,51 @@
...
+var invalidUser = require('../users').invalid;
+
+var validFile = require('./test-files').valid;
+var invalidFile = require('./test-files').invalid;
+
+var validProject = require('../projects').testProjects.valid;
+var invalidProject = require('../projects').testProjects.invalid;
+
+function generateUrl(opts) {
+  return '/user/' + opts.userid + '/projects/' + opts.projectid + '/files/' + opts.fileid;
```
gideonthomas: s/`/user`/`/users/`

gideonthomas:
looks good so far!

package.json
```diff
@@ -7,7 +7,8 @@
     "start": "node app",
     "env": "cp env.dist .env",
     "db": "knex migrate:latest && knex seed:run",
-    "test": "lab"
+    "test": "npm run test:setup && lab -t 100 -c --verbose --colors --assert code --timeout 5000",
+    "test:setup": "npm run test:droptables && npm run test:createtables && npm run test:insertdata",
```
sedge: This is incomplete. I was beginning to add this stuff, so don't expect it to work :sob:

ashleygwilliams:
you should be able to use `knex seed:run` to populate the db with test data, is there a reason you aren't doing that?

sedge:
No - I was copying from api. I'll use `knex seed:run` instead

ashleygwilliams:
that may eliminate the need to move the data out of the seeds...

sedge:
Moving the data out of the seeds was intended to eliminate the need to hardcode the tests with values found in the db. I figured it would be a win if they could both reference the same values.

ashleygwilliams:
ah ya, i see your point. :+1:

test/api/handlers/files/get-one.js
```diff
@@ -0,0 +1,96 @@
...
+});
+
+after(function(done) {
+  server.stop(done);
+});
+
+// GET /user/:userid/projects/:projectid/files/:fileid
+experiment('[Get one file]', function() {
+  experiment('[Parameter errors]', function() {
+    test('userid must be valid', function(done) {
```
ashleygwilliams: let's use underscores because that's what they are in the model? user_id

gideonthomas:
r+ merge away

PR #33, opened 2015-06-09
add new project attrs, update seeds

### LINKED ISSUE #29: Update data models to reflect the Teach team's requirements
Our data models need to evolve to identically map the UX workflow and representation of data on Thimble's side.

One of the recent changes that we discussed is to add a `tags` attribute to the projects model to store project-related tags that a user might specify. Another change was to add a `description` attribute to the project as well.

We probably eventually want to include thumbnails as well for projects depending on how the UX team decides to implement thumbnails.

ashleygwilliams:
fixes #29

### DIFF
--- a/migrations/1_create_users_projects_files.js
+++ b/migrations/1_create_users_projects_files.js
@@ -8,6 +8,10 @@ exports.up = function (knex) {
       t.increments('id');
       t.integer('user_id').notNullable().references('id').inTable('users');
       t.text('title').notNullable();
+      t.text('tags');
+      t.text('description');
+      t.text('date_created').notNullable();
+      t.text('date_updated').notNullable();
     })
     .createTable('files', function (t) {
       t.increments('id');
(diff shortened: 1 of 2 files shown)


ashleygwilliams:
lemme know what you think @gideonthomas

migrations/1_create_users_projects_files.js
```diff
@@ -8,6 +8,10 @@ exports.up = function (knex) {
       t.increments('id');
       t.integer('user_id').notNullable().references('id').inTable('users');
       t.text('title').notNullable();
+      t.text('tags');
```
sedge: I guess we'll have to parse them ourselves? Maybe store as csv...

gideonthomas: maybe we want a separate table for this? That might make things even easier when the Thimble gallery is implemented (for e.g. filtering makes by tags)?

gideonthomas:
this looks good to me!

migrations/1_create_users_projects_files.js
ashleygwilliams: when we do this in pg it shouldn't be so bad. http://www.postgresql.org/docs/9.1/static/arrays.html

gideonthomas: yeah but it will completely break in dev if we use sqlite

PR #34, opened 2015-06-09
add lout

### LINKED ISSUE #5: Document our API
The API is complex, and the attributes of each schema need to be explained in simple language.

ashleygwilliams:
fixes #5

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,3 +54,8 @@ to enforce the [`mofo-style-guide`](https://github.com/MozillaFoundation/javascr
 - To run the style checker, use `npm run jscs`.
 - To run the hinter, use `npm run jshint`.
 - To run both, use `npm run lint`.
+
+## Documentation
+
+This project uses [`lout`](https://github.com/hapijs/lout) to automatically generate API documention. To view the docs, start the server and visit
+[`http://localhost:2015/docs`](http://localhost:2015/docs).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -45,3 +45,8 @@ $ npm start
 
 The server's log level can be set in the environment or the .env file using `LOG_LEVEL=*` with one of `fatal`, `error`, `warn`, `info`, `debug`, `trace`.
 If none is given `info` is used.
+
+## Documentation
+
+This project uses [`lout`](https://github.com/hapijs/lout) to automatically generate API documention. To view the docs, start the server and visit
+[`http://localhost:2015/docs`](http://localhost:2015/docs).
```
sedge: I wonder if there's an easy way to capture the docs and stuff them in the `readme.md`?

ashleygwilliams: i feel like having the docs in the readme might be too much... i do think we could put a quick version of the available routes in their though, since they are p conventional.

sedge: You're probably right. Having to run the server to reference stuff just feels gross to me -.-

ashleygwilliams: i feel that. once we deploy tho, it will be hosted, no?

sedge: Good point! I hadn't thought of that.

sedge:
r+, lets get this in.

ashleygwilliams:
ok rebasing and squashing

Issue #25, opened 2015-06-08
Implement jshinting + travis-ci integration


sedge:
This could be two separate issues but they play well together.

simonwex:
This should be the mofo-style set of linters

`npm install mofo-style --save-dev`

https://github.com/MozillaFoundation/javascript-style-guide

ashleygwilliams:
solved by #35

PR #35, opened 2015-06-09
add jscs and mofo-style, npm script

ashleygwilliams:
solves #25

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
     "start": "node server",
     "env": "cp env.dist .env",
     "db": "knex migrate:latest && knex seed:run",
-    "test": "lab"
+    "test": "lab",
+    "jscs": "jscs -c node_modules/mofo-style/linters/.jscsrc api server.js"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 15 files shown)


ashleygwilliams:
so jshint barfs about const. how do we feel about this? cc/ @simonwex

simonwex:
Happy to move forward with ES6 in our style-guide. Timing might be an issue as @gvn and @Pomax who care greatly about this stuff are in :shipit: <(SHIP IT) mode.

sedge:
Shall we pull out `const` for now then?

package.json
```diff
@@ -7,7 +7,10 @@
     "start": "node server",
     "env": "cp env.dist .env",
     "db": "knex migrate:latest && knex seed:run",
-    "test": "lab"
+    "test": "lab",
```
gideonthomas: I think we should add `npm run lint` to the test script as well

sedge: I'll add it in #32

ashleygwilliams: :+1:

ashleygwilliams:
ok, playing nice with mofo-style. final review appreciated :bow:

gideonthomas:
r=me

ashleygwilliams:
ok squashing

PR #38, opened 2015-06-10
Added missing files to lint script and fixed lint errors

sedge:


ashleygwilliams:
lgtm :+1:

authentication/scheme.js
```diff
@@ -2,7 +2,7 @@
  * Authentication scheme used to authenticate requests to publish.webmaker.org
 */
 
-var Boom = require('boom');
+// var Boom = require('boom');
```
ashleygwilliams: why comment out?

sedge: Right? This file uses `boom` in its example code below, so I figured I'd leave this to make it clear that it's a dependency.

sedge: Oh, and jshint complains if I don't cuz the variable isn't used anywhere.

PR #40, opened 2015-06-10
[REVIEW] project file and files GET routes

### LINKED ISSUE #15: Implement GET /users/:userID/projects/:projectID/files/:fileID w/ tests
Returns the contents of a file.

ashleygwilliams:
fixes #15

do we need /users/id/ or can we just use project/id since ids are unique universally, not per user

### DIFF
--- a/api/modules/files/routes.js
+++ b/api/modules/files/routes.js
@@ -13,6 +13,18 @@ exports.register = function(server, options, next) {
     config: {
       handler: controller.getFile
     }
+  }, {
+    method: 'GET',
+    path: '/projects/{project_id}/files',
+    config: {
+      handler: controller.getProjectFiles
+    }
+  }, {
+    method: 'GET',
+    path: '/projects/{project_id}/files/{id}',
+    config: {
+      handler: controller.getProjectFile
+    }
   }]);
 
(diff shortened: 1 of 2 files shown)


ashleygwilliams:
@gideonthomas particularly interested in your thought on this. not including the user lookup makes the query much shorter

gideonthomas:
agree! I think we should do this for the `/files` routes as well and maybe also implement your filter idea

ashleygwilliams:
:+1:

api/modules/files/routes.js
```diff
@@ -13,6 +13,18 @@ exports.register = function(server, options, next) {
...
     }
+  }, {
+    method: 'GET',
+    path: '/projects/{project_id}/files',
+    config: {
+      handler: controller.getProjectFiles
+    }
+  }, {
+    method: 'GET',
+    path: '/projects/{project_id}/files/{id}',
```
gideonthomas: `/files/{id}`?

sedge: I thought we decided against this in channel?

gideonthomas:
@ashleygwilliams this is good to merge

PR #42, opened 2015-06-10
[REVIEW] add POST, PUT, DELETE routes and handlers for users

ashleygwilliams:
apparently there were not issues for these routes, but that's all good.

to test:

POST: `curl --data "name=sedge" http://localhost:2015/users`
PUT: `curl -X PUT --data "name=gideon" http://localhost:2015/users/4`
DELETE: `curl -X DELETE "name=gideon" http://localhost:2015/users/4`

check the browser at route `http:/localhost:2015/users` to see that the behavior is correct

### DIFF
--- a/api/modules/users/controller.js
+++ b/api/modules/users/controller.js
@@ -5,7 +5,34 @@ module.exports = {
     return reply(User.fetchAll());
   },
   getUser: function(req, reply) {
-    return reply(User.query('where', 'id', '=', req.params.id)
-      .fetch());
+    return reply(User.query({
+      where: {
+        id: req.params.id
+      }
+    }).fetch());
+  },
+  createUser: function(req, reply) {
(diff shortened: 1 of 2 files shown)


gideonthomas:
I don't know if we want to hold off on this patch for now or not until we decide on an authorization strategy because it seems like all we're using the Users table for is authorization. And from the client perspective, figuring out whether to use POST or GET (Thimble won't know if a signed in user is new to Thimble or not) might be a problem

ashleygwilliams:
is there a reason a Thimble user doesn't know if the user is new to Thimble?

sedge:
Thimble doesn't have a `user` db.

gideonthomas:
we can decide on how to change this later!
@ashleygwilliams this is also good to land!

Issue #16, opened 2015-06-08
Implement DELETE projects/:projectID/files/:fileID , files/:fileID


sedge:
Delete the given file from the working version of the project (not the published one)

ashleygwilliams:
there will be 2 routes to delete files:

```
DELETE projects/:project_id/files/:file_id
DELETE files/:file_id
```

ashleygwilliams:
this is solved by #44

PR #44, opened 2015-06-10
add POST, PUT, DELETE routes and handlers for files

ashleygwilliams:
solves #16 among adding other functionality

to test

POST: `curl --data "path=poop.com&project_id=1&buffer=XXX" http://localhost:2015/files`
PUT: `curl -X PUT --data "path=hello.com&project_id=1&buffer=XXX" http://localhost:2015/files/7`
DELETE: `curl -X DELETE  http://localhost:2015/files/7`

watch `http://localhost:2015/files` to see correct response

### DIFF
--- a/api/modules/files/controller.js
+++ b/api/modules/files/controller.js
@@ -25,5 +25,33 @@ module.exports = {
         id: req.params.id
       }
     }).fetch());
+  },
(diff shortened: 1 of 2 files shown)


gideonthomas:
nice! This works! r+
merge away :)

PR #46, opened 2015-06-11
Add sequence diagrams

gideonthomas:
I made a script to generate the sequence diagrams based on text files containing data syntactically based on the `websequencediagrams` interface. I did this so that henceforth, others can also update the diagrams easily.


ashleygwilliams:
this is *real* cool, @gideonthomas. great work. r+

gideonthomas:
thanks!
@simonwex would you like to take a look as well?

gideonthomas:
It seems like we're not going to be using this anytime soon, closing!

PR #51, opened 2015-06-11
add query for name on GET /users

### LINKED ISSUE #30: Implement GET /users?name=:username
This route returns a user object which will look something like:
```javascript
{
  id: ...,
  name: ...,
}
```

ashleygwilliams:
i did the easy thing and just put it on the single route, could improve later

fixes #30

### DIFF
--- a/api/modules/users/controller.js
+++ b/api/modules/users/controller.js
@@ -2,6 +2,13 @@ var User = require('./model.js');
 
 module.exports = {
   getUsers: function(req, reply) {
+    if(req.query.name) {
+      return reply(User.query({
+        where: {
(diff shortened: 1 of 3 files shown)


api/modules/users/controller.js
```diff
@@ -2,6 +2,14 @@ var User = require('./model.js');
 
 module.exports = {
   getUsers: function(req, reply) {
+    console.log(req.query.name);
```
sedge: Nuke this

ashleygwilliams: lol oops

sedge:
r+, merge away

PR #55, opened 2015-06-13
update to use postgres

ashleygwilliams:


knexfile.js
```diff
@@ -1,14 +1,16 @@
+var path = require('path');
+
 module.exports = {
   development: {
-    client: 'sqlite3',
-    debug: true,
+    client: 'pg',
+    debug: process.env.NODE_ENV !== 'production',
```
sedge: This is a great idea.

knexfile.js
```diff
@@ -1,14 +1,16 @@
...
     connection: {
```
sedge: These values should probably be environment variables.

package.json
```diff
@@ -6,8 +6,10 @@
   "scripts": {
     "start": "node server",
     "env": "cp env.dist .env",
-    "db": "knex migrate:latest && knex seed:run",
-    "test": "rm dev.db && npm run db && lab -t 100 -c --verbose --colors --assert code --timeout 5000 && npm run lint",
+    "knex": "knex migrate:latest && knex seed:run",
+    "db:new": "createdb publish && npm run knex",
```
sedge: You should remove the `createdb` and `dropdb` commands from here since a user can arbitrarily name their database now. What about an example in the README instead?

ashleygwilliams:
switching this to WIP... need to change up somethings for deploy

package.json
ashleygwilliams: why do we want to afford the user to make an arbitrary database name?

sedge: It gives devops control of this value, rather than us. This is [good design](http://12factor.net/config).

sedge:
r+

PR #59, opened 2015-06-15
Made knexfile load the env file during seeding

sedge:


ashleygwilliams:
my bad. thanks for catching this. r+

PR #62, opened 2015-06-15
Added cascading deletes and cascading drops on reseeding the database

sedge:


knexfile.js
```diff
@@ -8,7 +8,7 @@ if (!process.env.NODE_ENV) {
 module.exports = {
   development: {
     client: 'pg',
-    debug: process.env.DEBUG,
+    debug: process.env.DEBUG == true,
```
ashleygwilliams: do we want this in this commit?

ashleygwilliams:
except for my one question i am r+ :8ball:

knexfile.js
sedge: I'll separate it.

PR #67, opened 2015-06-16
implement /projects/:id/publish and /projects/:id/unpublish routes

ashleygwilliams:


api/classes/publish.js
```diff
@@ -0,0 +1,29 @@
+var AWS = require('aws-sdk'); 
+var s3 = new AWS.S3(); 
+
+var isPublished = function(project){
+  return project.get('publish_url') !== null;
+};
+
+exports.publish = function(project) {
+  var url = "FAKE://URL";
```
sedge: I assume this is where the s3 logic will go when the time comes?

api/classes/publish.js
```diff
@@ -0,0 +1,29 @@
...
+
+exports.unpublish = function(project) {
+  var message = "";
+  if (isPublished(project)) {
+    project.set({ publish_url: null }).save();
+    message =  "Successfully unpublished " + project.get('title');
+  } else {
+    message = "Cannot unpublish an project that is not published.";
+  }
+  console.log(message);
```
sedge: nuke this sucker

api/classes/publish.js
```diff
@@ -0,0 +1,29 @@
...
+var s3 = new AWS.S3(); 
+
+var isPublished = function(project){
+  return project.get('publish_url') !== null;
+};
+
+exports.publish = function(project) {
+  var url = "FAKE://URL";
+  project.set({ publish_url:  url }).save();
+  console.log("Successfully published " +
```
sedge: We have a logger don't we? Let's use that instead.

api/classes/publish.js
ashleygwilliams: indeed.

api/classes/publish.js
ashleygwilliams: oh right. lol. yes, that makes sense.

test/api/handlers/projects/publish.js
```diff
@@ -0,0 +1,68 @@
+var Lab = require('lab');
+var lab = exports.lab = Lab.script();
+
+var experiment = lab.experiment;
+var test = lab.test;
+var before = lab.before;
+var after = lab.after;
+var expect = require('code').expect;
+
+var config = require("../../../lib/fixtures/files").update;
```
sedge: This is supposed to contain the fail or success options being passed to the server in each test. This is pulling from the `files` fixtures, so these tests will behave strangely.

test/api/handlers/projects/unpublish.js
```diff
@@ -0,0 +1,68 @@
+var Lab = require('lab');
+var lab = exports.lab = Lab.script();
+
+var experiment = lab.experiment;
+var test = lab.test;
+var before = lab.before;
+var after = lab.after;
+var expect = require('code').expect;
+
+var config = require("../../../lib/fixtures/files").update;
```
sedge: ditto.

PR #70, opened 2015-06-16
Fix some create routes to pass tests

sedge:
This ended up fixing:

*  `POST /files`
*  `DELETE  /files/:file_id`
*  `GET /projects/:project_id/files`


sedge:
@ashleygwilliams I'm going to merge this in, since the project tests will rely on it.

PR #71, opened 2015-06-16
implement /projects/:id/publish and /projects/:id/unpublish routes

ashleygwilliams:


test/lib/fixtures/projects/unpublish.js
```diff
@@ -0,0 +1,37 @@
...
+    unpublish.success = {
+      default: {
+        url: '/projects/' + validProjects[0].id,
+        method: 'POST'
+      }
+    };
+
+    unpublish.fail = {
+      projectDoesNotExist: {
+        url: '/projects/999999',
```
sedge: `/projects/999999/unpublish`

test/lib/fixtures/projects/unpublish.js
```diff
@@ -0,0 +1,37 @@
...
+        method: 'POST'
+      },
+      projectidTypeError: {
+        url: '/projects/thisisastring',
```
sedge: ditto

test/lib/fixtures/projects/unpublish.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  retrieveTestProjects(function(err, projects) {
+    if (err) return cb(err);
+
+    validProjects = projects.valid;
+
+    unpublish.success = {
+      default: {
+        url: '/projects/' + validProjects[0].id,
```
sedge: ah, and here.

Also, keep in mind that this will rely on this test running after the publish test does. Otherwise, you're better off publishing and then unpublishing in the same test.

test/lib/fixtures/projects/unpublish.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  retrieveTestProjects(function(err, projects) {
+    if (err) return cb(err);
+
+    validProjects = projects.valid;
+
+    unpublish.success = {
+      default: {
+        url: '/projects/' + validProjects[0].id + '/unpublish',
```
sedge: My old comment:

ah, and here.

Also, keep in mind that this will rely on this test running after the publish test does. Otherwise, you're better off publishing and then unpublishing in the same test.

ashleygwilliams:
closing in favor of #76

PR #72, opened 2015-06-17
Finished fixing Files to pass tests

sedge:


api/modules/files/controller.js
```diff
@@ -1,27 +1,59 @@
...
+        id: req.params.project_id
       }
-    }).fetchAll()
+    }).fetch()
+    .then(function(record) {
+      if (!record) {
+        return Promise.reject('Project reference does not exist.');
+      }
+
+      return File.query({
```
ashleygwilliams: i don't think we want to do 2 queries here. also shouldn't we `.then` and `.catch` here?

sedge: By returning this promise, the `.then` and `.catch` on lines 54 and 61 will do the trick.

PR #76, opened 2015-06-18
kitchen sink

ashleygwilliams:
instead of #71 

currently at 24 of 70 failing

- refactor for all resources to use base_controller CRUD methods
- add publish/unpublish routes, handlers, and publisher class
- fix bugs in tests
- create error class, use to validate id param and payload on routes

missing tests, full s3 implementation for publish/unpublish (just a shell right now waiting on @jbuck )

### DIFF
--- a/api/modules/files/controller.js
+++ b/api/modules/files/controller.js
@@ -1,8 +1,5 @@
 var Boom = require('boom');
-var Promise = require('bluebird');
-
 var Model = require('./model');
-var Project = require('../projects/model');
 
 var BaseController = require('../../classes/base_controller');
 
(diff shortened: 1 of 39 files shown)


api/modules/files/controller.js
```diff
@@ -1,171 +1,49 @@
...
-        }
-      }
+controller.getProjectFiles = function(req, reply) {
+  var result =  Model.query({
+    where: {
+      project_id: req.params.project_id
+    }
+  }).fetchAll()
+  .then(function(records) {
+    if (!records) { throw Boom.notFound(); }
```
sedge: Why not `return reply(Boom.notFound())`?

sedge: Forget it, just read two lines down :P

sedge: Though I am curious about why this works...

ashleygwilliams: yeah, i actually got this straight from the Hapi maintainer: 

@nlf says: "hapi will resolve a promise for you if you reply with it, including catching errors, so rather than having multiple calls to reply() in your then and catch methods you can just do something like this: https://gist.github.com/nlf/be073ef93c3ce0918055"

Issue #77, opened 2015-06-18
#76: create doesn't respect `id` increments


ashleygwilliams:
CLIENT:
```
$ curl -X POST --data "name=poop" http://localhost:2015/users
{"statusCode":500,"error":"Internal Server Error","message":"An internal server error occurred"}
```

SERVER:
```
[2015-06-18T04:22:57.246Z] ERROR: publish.webmaker.org/67140
on Ashleys-MacBook-Pro.local: insert into "users" ("name") values
($1) returning "id" - duplicate key value violates unique constraint "users_pkey"
```

on a subsequent `POST` the `id` increments, eventually works once the `id` > 3 (after 3 tries)

sedge:
Bookshelf threw the following error message (along with other data): `"insert into "users" ("name") values ($1) returning "id" - duplicate key value violates unique constraint "users_pkey""`

Do we have a constraint on `name` forcing it to be unique?

sedge:
I see now there's not much new to go off of here. I'll dig into the error object a little more.

sedge:
@ashleygwilliams At least for the user, I know what the problem is. The seeds declare the ids for the first three users manually, which must not be auto-incrementing the counter in the DB. I'll throw up a PR for that.

PR #81, opened 2015-06-18
pass tests

ashleygwilliams:
READ, UPDATE, and DELETE functionality totally works, the tests don't pass tho

CREATE still has issue #77

### DIFF
--- a/api/classes/base_controller.js
+++ b/api/classes/base_controller.js
@@ -41,21 +41,24 @@ BaseController.prototype.create = function(req, reply) {
 };
 
 BaseController.prototype.update = function(req, reply) {
+  var reqData = this.data(req);
   var result = this.Model.query({
     where: {
       id: req.params.id
     }
   })
-  .save(this.data(req), {
-    method: 'update',
-    patch: 'true',
-    require: 'true'
-  })
-  .call('fetch')
+  .fetch({ require: true })
   .then(function(record) {
     if (!record) { throw Boom.notFound(); }
-    return req.generateResponse(record.toJSON())
-      .code(201);
+    record.set(reqData);
+    if (!record.hasChanged()) {
+      throw Boom.create(500, 'Update failed');
+    }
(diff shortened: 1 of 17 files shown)


ashleygwilliams:
currently at 20 of 70 failing

ashleygwilliams:
17 of 70

api/classes/base_controller.js
```diff
@@ -41,37 +41,25 @@ BaseController.prototype.create = function(req, reply) {
...
+      throw Boom.create(404);
     }
+    return record
+      .save(record.changed, { patch: true, method: 'update' });
   })
-  .save(this.data(req), {
-    method: 'update',
-    patch: 'true'
-  });
-  var updated_state = this.Model.query({
```
sedge: I have no idea how this worked before. All this stuff happening in parallel must have been a nightmare to debug.

api/modules/files/controller.js
```diff
@@ -12,7 +12,7 @@ controller.data = function(req) {
     buffer: req.payload.buffer
   };
   if (req.params.id) {
-    data.id = req.params.id;
+    data.id = parseInt(req.params.id);
```
sedge: Strange that you have to do this. It's an odd place for this to be forced. Why not add param validation with Joi?

api/modules/files/routes/update.js
```diff
@@ -8,7 +10,11 @@ module.exports = [{
     handler: controller.update.bind(controller),
     description: 'Update a single file object based on `id`.',
     validate: {
-      payload: schema
+      payload: schema,
+      params: {
```
sedge: Huh. To my point above, why isn't this enough?

sedge: Could add to `create` too?

ashleygwilliams:
16 of 70

sedge:
Closed in favour of #87

PR #84, opened 2015-06-18
Fix #82 - Correct buffer handling

### LINKED ISSUE #82: Change validation for the `buffer` attribute of files before persisting to the db
In order to support sending buffers as JSON, we will be receiving the octets of the buffer as an array. So first, the Joi payload validation must check for an array vs a buffer, and second we must try converting that array into a node buffer. If the conversion succeeds, we can persist it to the DB. Otherwise, it's a malformed request.

sedge:


sedge:
@ashleygwilliams r?

PR #91, opened 2015-06-19
fix create and login

ashleygwilliams:


api/classes/base_controller.js
```diff
@@ -38,8 +38,7 @@ BaseController.prototype.create = function(req, reply) {
         .code(201);
     })
     .catch(function(e) {
-      // We might want to parse the error types here.
-      reply(e);
+      throw Boom.wrap(e);
```
sedge: Can we `reply` instead of throw?

sedge: Rather, why are we throwing?

PR #96, opened 2015-06-29
[REVIEW] publish/unpublish to s3

ashleygwilliams:


ashleygwilliams:
i need a cloudflare/amazon account to see if this is working correctly

sedge:
@jbuck Can you co-ordinate with @ashleygwilliams for this? ^^

api/classes/publisher.js
```diff
@@ -1,19 +1,108 @@
...
+var msg_success = function(project) { 
+  log.info( 'Successfully published ' +
+            project.get('title') +
+            ' to ' + project.get('publish_url'));
+  
+  return project;
+};
+
+exports.publish = function(project) {
+  new Promise(
```
gideonthomas: needs to be required at the top + package.json entry

api/classes/publisher.js
```diff
@@ -1,19 +1,108 @@
...
+      } else {
+        log.info(res.statusCode);
+        log.info('Successfully unpublished ' + project.get('title'));
+      }
+    })
+  }).catch(function(e) {
+    log.error(e);
+    return false;
+  });
+  return true;
```
gideonthomas: I'm pretty sure this should be inside the `.then` callback, otherwise it will always return `true`.

api/classes/publisher.js
gideonthomas: Actually, you don't need to use this once `publishFiles` returns a promise, just need to attach a `.then` and `.catch` callback

api/classes/publisher.js
```diff
@@ -1,19 +1,108 @@
...
+  }).fetchAll()
+  .then(function(records){
+    records.mapThen(function(record) {
+      uploadFile(record, params);
+    });
+  }).catch(function(e) {
+    log.error(e);
+    return false;
+  });
+  return true;
```
sedge: This will always return true, since the `return false` on line 27 is inside a callback function.

Maybe try something like:

```js
return File.query(...)
  .then(fn() {
    // ... 
    return Promise.resolve(true); 
  }).catch(fn(e) {
    return Promise.reject(false);
  });
```

api/classes/publisher.js
```diff
@@ -1,19 +1,108 @@
...
+  var key = file.get('path');  
+  var headers = {
+    'Content-Type': 'application/octet-stream'
+  }
+
+  client.putBuffer(buffer, key, headers, function(err, res){
+    if(err) {
+      log.error(err)
+    } else {
+      if (200 == res.statusCode) {
```
sedge: Nit: `===`

api/classes/publisher.js
sedge: Ah, Gideon caught it too. See my comment above for a possible solution.

api/classes/publisher.js
```diff
@@ -1,19 +1,108 @@
...
+    })
+  }).catch(function(e) {
+    log.error(e);
+    return false;
+  });
+  return true;
+}
+
+var build_url= function(project) {
+  var url = 'd3g6se91hcs80r.cloudfront.net' + '/' +
```
sedge: This URL should be in an environment var

api/classes/publisher.js
```diff
@@ -1,19 +1,108 @@
...
+      }
+    })
+  }).catch(function(e) {
+    log.error(e);
+    return false;
+  });
+  return true;
+}
+
+var build_url= function(project) {
```
sedge: camelCase. I suspect the server is long overdue for a jshint run :P

api/classes/publisher.js
```diff
@@ -1,19 +1,108 @@
...
+  var url = 'd3g6se91hcs80r.cloudfront.net' + '/' +
+             client.bucket + '/' +
+             project.get('title');
+  return url;
+}
+
+var set_url = function(project, url){
```
sedge: Nit: camelCase

api/classes/publisher.js
sedge: +1

api/classes/publisher.js
```diff
@@ -1,19 +1,108 @@
...
   project.set({ publish_url:  url }).save();
-  log.info('Successfully published ' +
-          project.get('title') +
-          ' to ' + project.get('publish_url'));
   return project;
+}
+
+var msg_success = function(project) {
```
sedge: Nitpicking hat trick: camelCase

api/classes/publisher.js
```diff
@@ -1,19 +1,108 @@
...
-exports.publish = function(project) {
-  var url = 'FAKE://URL';
+var publishFiles = function(project, params) {
+  File.query({
+    where: {
+      project_id: project.id
+    }
+  }).fetchAll()
+  .then(function(records){
+    records.mapThen(function(record) {
```
sedge: This should be:

```js
.then(fn(record) {
  return records.mapThen(...);
}).then(fn() {
  // All records processed 
}).catch(...);
```

api/classes/publisher.js
```diff
@@ -1,19 +1,108 @@
...
+}
+
+var deleteFiles = function(project) {
+  File.query({
+    where: {
+      project_id: project.id
+    }
+  }).fetchAll()
+  .then(function(records){
+    client.deleteMultiple(records.models, function(err, res){
```
sedge: This will probably need to be promise-based vs. callback so you can continue the chain:

```js
.fetchAll()
.then(fn(records) {
  return client.deleteMultiple(records.models);
})
.then(fn() {
  // all deleted successfully
  return true;
})
.catch(...)

**Comment by sedge** [comment]
@ashleygwilliams A few pieces (mostly promise related) need to be fixed, and I picked a few nits here and there. Feel free to ignore those haha

File: api/classes/publisher.js [hunk 11]
```diff
@@ -1,31 +1,126 @@
...
+  secret: process.env.AWS_SECRET_ACCESS_KEY,
+  bucket: process.env.AWS_BUCKET
+});
 
 var log = require('../../logger.js');
+var File = require('../modules/Files/model');
 
-var isPublished = function(project) {
-  return project.get('publish_url') !== null;
+function publishFiles(project, params) {
```
gideonthomas: Where do you send `params`?

api/modules/projects/controller.js
```diff
@@ -70,8 +83,18 @@ controller.unpublishProject = function(req, reply) {
   }).fetch()
   .then(function(record) {
     if (!record) { throw Boom.notFound(); }
-    return req.generateResponse(Publisher.unpublish(record))
-      .code(201);
+
+    return Publisher.unpublish(record)
+      .then(function() {
+        return req.generateResponse(record).code(201);
```
gideonthomas: HTTP 200

api/modules/projects/controller.js
```diff
@@ -70,8 +83,18 @@ controller.unpublishProject = function(req, reply) {
...
+      });
+  })
+  .catch(function(e) {
+    if (e.isBoom) {
+      return e;
+    }
+
+    return Boom.badImplementation(e);
   });
   return reply(result);
```
gideonthomas: blank line before this

api/classes/publisher.js
```diff
@@ -1,31 +1,129 @@
...
+    .then(function() {
+      return project.set({ publish_url: null }).save();
+    })
+    .then(function() {
+      log.info(
+        'Successfully unpublished ' +
+        project.get('title')
+      );
+
+      return Promise.resolve();
```
gideonthomas: You don't need this

api/classes/publisher.js
```diff
@@ -1,31 +1,129 @@
...
+
+    log.info('Successfully uploaded ' + key);
+    deferred.resolve();
+  });
+
+  return deferred;
+}
+
+function deleteFiles(project) {
+  File.query({
```
gideonthomas: return

api/modules/projects/controller.js
```diff
@@ -3,6 +3,7 @@ var Model = require('./model');
 var BaseController = require('../../classes/base_controller');
 var Publisher = require('../../classes/publisher');
 var controller = new BaseController(Model);
+var logger = require('../../../logger.js');
```
gideonthomas: `require('../../../logger');`

api/classes/publisher.js
```diff
@@ -1,31 +1,143 @@
...
+      return fn(record, project.get('user_id'));
+    });
+  })
+  .catch(function(e) {
+    return Promise.reject(e);
+  });
 };
 
+function uploadFile(file, userId) {
+  var deferred = new Promise();
```
gideonthomas: let's not call this `deferred` and instead call it `promise` coz deferred has a different meaning (especially in jquery)

sedge: This is a convention I've seen in a few different places. Isn't a jquery `deferred` object a promise anyway?

gideonthomas: Not exactly, it's more of a wrapper! To return a promise you have to do `deferred.promise()` in jquery

sedge:
@gideonthomas r+'d in channel

PR #114, opened 2015-07-15
Fix #113 - Store an exact copy of the published files in a project

### LINKED ISSUE #113: Store a copy of published projects in a separate database
For this to be useful for upcoming work (injecting html for remix, rename cases) it should be decoupled from other manipulation of the project resource.

I'm imagining a separate table exclusively for published versions of projects.

sedge:


api/classes/publisher.js
```diff
@@ -30,6 +37,10 @@ function buildUrl(project, user) {
   return url;
 }
 
+function getUser(project) {
```
gideonthomas: Why does this show up as a change?

sedge: Because I moved it up in the file.

gideonthomas: Ah I see, coz you moved it

api/classes/publisher.js
```diff
@@ -98,53 +105,133 @@ function modifyFiles(action, project, user) {
...
+    // Then, delete them from the DB
+    var deletePublishedProject = publishedProject.destroy();
+    var deletePublishedFiles = Promise.map(publishedFiles, function(file) {
+      return file.destroy();
     });
-};
 
-exports.unpublish = function(project) {
+    return Promise.join(deletePublishedFiles, deletePublishedProject, function() {
+      // We deleted things!
```
gideonthomas: What are we doing here?

api/classes/publisher.js
```diff
@@ -98,53 +105,133 @@ function modifyFiles(action, project, user) {
...
-    })
-    .catch(function(e) {
-      log.error(e);
-
-      return Promise.reject(e);
+  .then(function(u) {
+    user = u;
+
+    return getPublishedProject(project)
+    .then(function(oldPublishedProject) {
```
gideonthomas: why are you attaching it here instead of chaining it to the then on line 163

api/classes/publisher.js
```diff
@@ -98,53 +105,133 @@ function modifyFiles(action, project, user) {
...
+        project_id: project.id
+      }
+    }).fetchAll()
+    .then(function(files) {
+      if (files.length === 0) {
+        throw 'No files to be published!';
+      }
+
+      return files.mapThen(function(file) {
+        return PublishedFiles.forge({
```
gideonthomas: This will fail when updating published files no? You're **always** creating new ones.

gideonthomas: Oh you're deleting the old one, got it!

api/modules/projects/model.js
```diff
@@ -7,6 +7,9 @@ var instanceProps = {
   },
   files: function() {
     return this.hasMany(require('../files/model'));
+  },
+  publishedFiles: function() {
+    return this.hasMany(require('../publishedFiles/model'));
```
gideonthomas: Why are we maintaining a relationship between publishedFiles and projects?

api/classes/publisher.js
sedge: Promise.join returns a promise that resolves to whatever the callback function returns. The callback function is only executed when the promises we're joining resolve successfully, so once the callback runs we know that the project and the files were deleted. Returning undefined is fine, since all it does is signal to the next `.then()` block that the deletions are done.

api/classes/publisher.js
sedge: We never update published files. destroy + create

api/modules/publishedProjects/model.js
```diff
@@ -0,0 +1,31 @@
+var BaseModel = require('../../classes/base_model');
+
```
gideonthomas: remove space

api/modules/publishedProjects/model.js
```diff
@@ -0,0 +1,31 @@
...
+var Projects = require('../projects/model');
+
+var instanceProps = {
+  tableName: 'publishedProject',
```
gideonthomas: nit: as a general convention in sql (I don't know about postgres), table names are usually `_` delimited and not camel-cased

api/classes/publisher.js
sedge: Since I cache the user I *could* do what you're suggesting. If I wasn't, this would keep the `u` variable in scope for the `clearPublishedProject()` call on line 169. I'm happy to make that change.

api/modules/projects/model.js
sedge: Good catch - this is an artifact. *deletes*

api/modules/publishedProjects/model.js
sedge: Correct - we can change this if you want. The reason that's a convention is because sql has traditionally been really unreliable with case sensitivity. In postgres, it will respect it.

Your call!

api/classes/publisher.js
gideonthomas: OK

api/classes/publisher.js
gideonthomas: Yep would prefer if you changed it so that we're consistent

api/modules/publishedProjects/model.js
gideonthomas: It's fine we can leave it like this

api/modules/publishedProjects/model.js
sedge: Nah, this follows the convention I set up in other model files.

api/classes/publisher.js
```diff
@@ -79,72 +90,151 @@ function deleteFile(file, root) {
...
-function getUser(project) {
-  return project.user().query({}).fetch();
+function getPublishedProject(project) {
+  return PublishedProject.query({
+    where: {
+      project_id: project.get('id')
+    }
+  }).fetch()
+  .then(function(publishedProject) {
+    if (!publishedProject) {
```
gideonthomas: just `return publishedProject` it will return `undefined` anyway

api/classes/publisher.js
```diff
@@ -79,72 +90,151 @@ function deleteFile(file, root) {
...
-      );
-    })
-    .catch(function(e) {
-      log.error(e);
+  .then(function(u) {
+    user = u;
+  })
+  .then(function() {
+    return getPublishedProject(project)
+    .then(function(oldPublishedProject) {
```
gideonthomas: Make it have it's own scope

api/classes/publisher.js
```diff
@@ -79,72 +90,151 @@ function deleteFile(file, root) {
...
-      return project.set({ publish_url: null }).save();
-    })
-    .then(function() {
-      log.info(
-        'Successfully unpublished ' +
-        project.get('title')
-      );
+  .then(function(user) {
+    return getPublishedProject(project)
+    .then(function(publishedProject) {
```
gideonthomas: why not in its own scope?

gideonthomas:
few questions, otherwise this looks good!

api/classes/publisher.js
sedge: Because it needs access to the `user` parameter

api/classes/publisher.js
```diff
@@ -79,72 +90,151 @@ function deleteFile(file, root) {
...
+      'Successfully published ' +
+      project.get('title') +
+      ' to ' + project.get('publish_url')
+    );
+  })
+  .catch(reject);
 };
 
-exports.unpublish = function(project) {
+exports.unpublish = function unpublish(project) {debugger;
```
sedge: forgotten debugger :(

api/classes/publisher.js
gideonthomas: So lets do one or the other, either cache the user and keep the promise structure flat like you do in publish or follow this structure in publish too

gideonthomas:
Actually, I do have a pretty big concern about this patch and it has to do with republishing by unpublish followed by a publish. The way you have it, you delete the old records for the published project and generate new records in the db. I don't think that's such a good idea because it's not scalable. Because everytime you hit Publish, I delete the published files with ID 200, 201, 202 and then recreate the files with IDs 203, 204, 205. This means that we are heavily relying on very very extensible indexing by the db as this value can easily go to 100000000+. Instead, I think we should replace the contents of the old record with that of the new one so that the ID remains the same. Only on deletes should we remove the record entirely.

sedge:
You raise a good point. The best we could do is match paths, updating where possible. Otherwise we have to delete and republish to prevent old files from sticking around.

gideonthomas:
Gonna file a separate issue re: this

api/classes/publisher.js
```diff
@@ -30,6 +40,35 @@ function buildUrl(project, user) {
...
+  return project.user().query({}).fetch();
+}
+
+function getPublishedProject(project) {
+  return PublishedProject.query({
+    where: {
+      id: project.get('published_id')
+    }
+  }).fetch()
+  .then(function(publishedProject) {
```
gideonthomas: You don't need this `then` if all it does is return the parameter

sedge: +1

api/classes/publisher.js
```diff
@@ -79,72 +118,238 @@ function deleteFile(file, root) {
...
-    where: {
-      project_id: project.id
+function modifyPublishedFiles(action, project, user, files) {
+  return Promise.resolve()
+  .then(function() {
+    if (files.length) {
+      return files.mapThen(function(file) {
+        return action(file, '/' + user.get('name') + '/' + project.get('id'));
+      });
+    } else {
```
gideonthomas: No need for an `else`

api/classes/publisher.js
```diff
@@ -79,72 +118,238 @@ function deleteFile(file, root) {
...
-    return buildUrl(project, user);
+    return getPublishedProject(project)
+    .then(function(oldPublishedProject) {
+      if (oldPublishedProject) {
+        return oldPublishedProject.set({
+          title: project.get('title'),
+          tags: project.get('tags'),
+          description: project.get('description')
+        }).save();
+      } else {
```
gideonthomas: no need of `else`

api/classes/publisher.js
```diff
@@ -79,72 +118,238 @@ function deleteFile(file, root) {
...
+      if (f.length === 0) {
+        return;
+      }
+      publishedFiles = f;
+    });
+  })
+  .then(function() {
+    // We want to organize 'file' objects by id, and
+    // 'publishedFile' objects by the 'file' objects they reference
+    files.models.forEach(function(file) {
```
gideonthomas: Consider using this instead: http://bookshelfjs.org/#Collection-Underscore-Methods

api/classes/publisher.js
```diff
@@ -79,72 +115,236 @@ function deleteFile(file, root) {
...
+    publishedFiles.forEach(function(publishedFile) {
+      if (publishedFile.get('file_id') === null) {
+        toBeDeleted.push(publishedFile);
+        return;
+      }
+      publishedProjectFiles[publishedFile.get('file_id')] = publishedFile;
+    });
+    var publishedFileForeignKeys = Object.keys(publishedProjectFiles);
+
+    filePrimaryKeys.forEach(function(primaryKey) {
```
gideonthomas: Can you add a comment here about what exactly you're doing?

api/classes/publisher.js
```diff
@@ -79,72 +115,236 @@ function deleteFile(file, root) {
...
-    .then(function(user) {
-      return modifyFiles(uploadFile, project, user);
-    })
-    .then(function(url) {
-      return project.set({
-        publish_url: url
+    // Nothing published yet?
+    if (!publishedFiles) {
+      return filePrimaryKeys.forEach(function(id) {
+        return toBeCreated.push(projectFiles[id]);
```
gideonthomas: you shouldn't need to return here

api/classes/publisher.js
gideonthomas: Also, I wonder if you can separate them into functions? Maybe that would make it clearer! Otherwise it seems hard to follow!
e.g. `function projectDeleteDiffs()`, `function projectCreateDiffs()`... or something like that

api/classes/publisher.js
```diff
@@ -79,72 +115,236 @@ function deleteFile(file, root) {
...
+      // If the reference exists in a 'publishedFile' object as
+      // a foreign key, and in a 'file' object as a primary key,
+      // we will be forcing an update in case the file changed before
+      // republishing.
+      if (filePrimaryKeys.indexOf(foreignKey) !== -1) {
+        toBeUpdated.push(publishedProjectFiles[foreignKey]);
+      }
+    });
+  })
+  .then(function() {
```
gideonthomas: I think it's more clear this way when you have them as separate functions for create, update and delete

api/classes/publisher.js
```diff
@@ -79,72 +115,236 @@ function deleteFile(file, root) {
...
+    // Confirm we have files to be updated
+    if (toBeUpdated.length === 0) {
+      return;
+    }
+
+    return Promise.map(toBeUpdated, function(model) {
+      return modifyPublishedFiles(deleteFileRemotely, project, user, model)
```
gideonthomas: so why do you first delete -> create for an update instead of just update?

api/classes/publisher.js
```diff
@@ -79,72 +115,236 @@ function deleteFile(file, root) {
...
+      .then(function() {
+        // Then, remove the reference from the parent project
+        return project.set({
+          published_id: null
+        }).save();
+      })
+      .then(function() {
+        // Finally, delete them from the DB
+        var deletePublishedProject = publishedProject.destroy();
+        var deletePublishedFiles = files.mapThen(function(file) {
```
gideonthomas: Do you need to do this? if the publishedFiles has a FK to publishedProject with cascade on delete enabled, it should automatically delete them shouldn't it?

api/modules/files/controller.js
```diff
@@ -15,4 +18,32 @@ controller.data = function(req) {
...
+  .then(function(model) {
+    if (!model) {
+      return;
+    }
+    return model.set({
+      file_id: null
+    }).save();
+  })
+  .then(function(m) {
+    BaseController.prototype.delete.call(self, req, reply);
```
gideonthomas: You shouldn't have to do this because `BaseController`'s methods are set up in such a way that they user "self" **always**. And in the `delete` function, I don't see `this` being used anywhere

api/modules/publishedProjects/model.js
```diff
@@ -0,0 +1,31 @@
+var BaseModel = require('../../classes/base_model');
+
+var Projects = require('../projects/model');
+
+var instanceProps = {
+  tableName: 'publishedProjects',
+  project: function() {
+    return this.belongsTo(Projects);
```
gideonthomas: Doesn't this create an FK in the publishedProjects table for Projects? Shouldn't it be the other way around?

api/classes/publisher.js
sedge: `Path` might have changed, so we have to remove the old one in case.

gideonthomas: Ah ok got it!

api/modules/publishedProjects/model.js
sedge: This uses bookshelf to create more abstract relationships as a layer on top of the DB's formal schemas. This method creates a query that essentially says, *get me the `project` whose `published_id` value is this `publishedProjects`' `id`*

api/classes/publisher.js
sedge: I find jumping between control logic and named functions that aren't reused confusing, which is why I avoid it in my code. I know you use this technique and I try to love it but my brain just won't! 

I'll compromise by commenting each of the largest `.then()` blocks to make it clear what they're doing.

api/modules/publishedProjects/model.js
sedge: Which raises the point of why I don't use it anywhere. -.-

gideonthomas:
r+

PR #115, opened 2015-07-16
Fix #111 - Add username to publish url and use project id instead of name

### LINKED ISSUE #111: Add username to published url
This would switch the url from `HOSTNAME/id/project` to `HOSTNAME/username/project`

gideonthomas:


sedge:
r+

PR #119, opened 2015-07-22
Fix #52 - Use form/multi-part instead of JSON to send file data

### LINKED ISSUE #52: Don't send file data through JSON
otherwise we end up encoding/decode the buffer

here's a npm package to use: https://www.npmjs.com/package/protobufjs
here's why we should use it: http://blog.codeclimate.com/blog/2014/06/05/choose-protocol-buffers/

sedge:


api/classes/prerequisites.js
```diff
@@ -153,4 +154,24 @@ prerequisites.validateCreationPermission = function(foreignKey, model) {
   };
 };
 
+/**
+ * trackTemporaryFile()
+ *
+ * Stores paths to temporary files in req.app for clearing after a request completes
+ * and in req.pre for use in the handler
+ */
+prerequisites.trackTemporaryFile = function() {
```
gideonthomas: Files* vs. File

sedge: Actually, I think this is 'file' now. At least ever since I figured out wtf was up with that two file nonsense :P

server.js
```diff
@@ -49,6 +51,27 @@ server.register({ register: require('./api') }, function(err) {
     });
     throw err;
   }
+
+  server.ext('onPreResponse', function(req, reply) {
```
sedge: I think I should probably factor this out into another file so I can give it a name.

```js
server.ext('onPreResponse', clearTemporaryFiles);
```

or something...

api/modules/files/routes/create.js
```diff
@@ -9,7 +9,14 @@ module.exports = [{
   method: 'POST',
   path: '/files',
   config: {
+    payload: {
+      allow: 'multipart/form-data',
+      parse: true,
+      output: 'file',
+      maxBytes: 1048576 * 5 // 5mb
```
gideonthomas: nit: maybe make this an env var?

test/lib/fixtures/files/create.js
```diff
@@ -8,13 +8,35 @@ var validFiles;
...
-var testBuffer = (new Buffer('test')).toJSON().data;
-
-var userToken = {
-  authorization: 'token ag-dubs'
+var validHeaders = {
+  authorization: 'token ag-dubs',
+  'content-type': 'multipart/form-data; boundary=AaB03x'
 };
 
+function constructMultipartPayload(fields, data) {
```
sedge: I normally despise code duplication, but these tests really need an overhaul anyway so I got lazy.

api/classes/prerequisites.js
gideonthomas: Ah nvm

server.js
```diff
@@ -49,6 +51,27 @@ server.register({ register: require('./api') }, function(err) {
...
+      return reply.continue();
+    }
+
+    // Once a successful request completes, we delete any
+    // temporary files we created
+    req.response.once('finish', function() {
+      if (req.app.tmpFile) {
+        fs.unlink(req.app.tmpFile, function(err) {
+          if (err) {
+            throw 'Failed to destroy temporary file with ' + err;
```
gideonthomas: I don't think you should throw and crash the entire app

server.js
```diff
@@ -49,6 +51,27 @@ server.register({ register: require('./api') }, function(err) {
...
+
+  server.ext('onPreResponse', function(req, reply) {
+    if (req.response.isBoom) {
+      return reply.continue();
+    }
+
+    // Once a successful request completes, we delete any
+    // temporary files we created
+    req.response.once('finish', function() {
+      if (req.app.tmpFile) {
```
gideonthomas: `if(!req.app.tmpFile) { return; }` and then the rest of the logic

server.js
gideonthomas: +1

test/lib/fixtures/files/create.js
gideonthomas: I wouldn't worry about the test code too much!

server.js
sedge: Good catch. My real mistake here was not explicitly handling the error, since HAPI will create an error response when you have an unhandled exception in the request lifecycle.

server.js
sedge: +1

env.dist
```diff
@@ -32,3 +32,6 @@ export LOG_LEVEL="info"
 # Location of the webmaker identity provider server (id.webmaker.org)
 # Note: Make sure to include the uri protocol in the connection string
 export ID_SERVER_CONNECTION_STRING="http://localhost:1234"
+
+# File size upload limit (defaults to 5mb in bytes)
+export FILE_SIZE_LIMIT=5242880
```
gideonthomas: blank line

adaptors/extensions.js
```diff
@@ -0,0 +1,23 @@
...
+      if (err) {
+        req.log.error('Failed to destroy temporary file with ' + err);
+      }
+    });
+  });
+
+  reply.continue();
+};
+
+module.exports = exports;
```
gideonthomas: I'm pretty sure you don't need to do this

sedge:
@gideonthomas r+ in channel

PR #130, opened 2015-08-04
Fixed #94 - Made update operations return 200 if nothing was modified

### LINKED ISSUE #94: Update operations should return 200 if no data was changed
https://github.com/mozilla/publish.webmaker.org/blob/master/api/classes/base_controller.js#L58 is not a good idea. It makes it really hard for the client to differentiate between "The id I sent does not reference an existing resource" and "The data I sent had no real effect". The second case doesn't really imply an error case so a 404 might not be a good way to deal with it.

sedge:
Because the test suite is cumbersome to update, I added a single test case for this behaviour. Since all implementations lean on the `base_controller` update method, this is sufficient to prove that it behaves properly.

### DIFF
--- a/api/classes/base_controller.js
+++ b/api/classes/base_controller.js
@@ -36,7 +36,7 @@ BaseController.prototype.update = function(req, reply) {
 
     record.set(reqData);
     if (!record.hasChanged()) {
-      throw Boom.notFound();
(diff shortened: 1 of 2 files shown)


gideonthomas:
r+

PR #138, opened 2015-08-13
Fix remix metadata injection regex to look for optional newline after head tag

gideonthomas:


lib/remix.js
```diff
@@ -13,7 +13,7 @@ function injectMetadata(html, metadata) {
     metaTags += '<meta name="data-remix-' + key + '" content="' + metadata[key] + '">\n';
   });
 
-  return html.replace(/<head>\n/, '$&' + metaTags);
+  return html.replace(/<head>\n?/, '$&' + metaTags);
```
humphd: I'd simplify this and just jam it right next to the closing `>`, forget about new line:

```js
return html.replace(/<head>/, '$&' + metaTags);
```

What about a page with no `<head>`?  This is legal html5

```html
<!DOCTYPE html>
<title>Messing with you</title>
Body Content
```

@Pomax might have ideas on how to do this for all cases.  I think it's fine to do a "we assume `<head>`" fix now.

Pomax: that sounds like a solid assumption. There are a lot of weird but legal HTML5 documents possible, but for the purposes of something like brimble assuming `<head>` seems sensible. It can also be forced from the thimble side, so that if people try to publish to our endpoint without a `<head>...</head>` block, that's an error. Even if HTML5 allows lots of different legal forms, our publish.wmo can be stricter about what it expects.

humphd:
@gideonthomas r+ on either version, but I'd skip the `\n` check personally, and just put it beside the `>`.  Clean code be damned!

PR #147, opened 2015-09-10
Fix Content-Type for IE so decompression works for tarballs

humphd:
Same fix we did for Thimble in https://github.com/mozilla/thimble.webmaker.org/issues/1064, where we switched away from `application/x-tar` for the `content-type` on a tarball in favour of `application/octet-stream` so that IE will properly decompress the stream in the browser.

### DIFF
--- a/api/classes/base_controller.js
+++ b/api/classes/base_controller.js
@@ -107,8 +107,10 @@ BaseController.prototype.getAllAsTar = function(req, reply) {
     });
   }
 
+  // Normally this type would be application/x-tar, but IE refuses to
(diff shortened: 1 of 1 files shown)


sedge:
r+

humphd:
@sedge, thanks for the review.  Would you mind landing/shipping this to staging/prod for me?  I'm in meetings almost the whole day.

gideonthomas:
@sedge we don't want thimble to receive it as an `application/x-tar` and then send it to the browser as `application/octet-stream` so might as well be consistent.

PR #151, opened 2015-09-14
Fix #150 - Experiment with piping the tar stream as soon as it is initialized

### LINKED ISSUE #150: Improve efficiency of project tarball creation
We looked at newrelic data and it's clear that our [code to build tarballs](https://github.com/mozilla/publish.webmaker.org/blob/master/api/classes/base_controller.js#L86-L114) needs improving (separate to work we might also do with #148).  Specifically, we are creating the tarball in memory and waiting until the entire promise chain for getting all files completes before we respond, which will lead to timeouts.

A better solution would be for us to switch to https://www.npmjs.com/package/tar-stream and start piping the tar stream immediately, then get a cursor to the server and loop through all the buffers we need, adding them as `entry` objects to the tar stream.

We're going to do an experiment next week to see how much we an improve things with this approach.

gideonthomas:


api/classes/base_controller.js
```diff
@@ -85,31 +85,29 @@ BaseController.prototype.getAllAsMeta = function(req, reply) {
...
       });
     }).then(function() {
-      archive.finalize();
-
-      return archive;
+      tarStream.finalize();
     });
   }
 
+  startStreaming().catch(errors.generateErrorResponse);
```
humphd: `startStreaming(files).catch(..)`  Let's pass `files in as a local arg instead of accessing via closure.

humphd: actually maybe try passing in `self.Model` too.  I want to reduce the amount we hang onto via closure.

humphd:
This looks great to me.  Just modify `startStreaming` to take args instead of reaching outside to access via closure, and this is r+.

PR #154, opened 2015-09-17
Fixed #143 - Added useful error logs

### LINKED ISSUE #143: Improve logging
We need to improve our logging system to log every time we successfully send a request or fail somehow!

Flagging this as an important issue because we need to be able to debug https://github.com/mozilla/thimble.webmaker.org/issues/1033

sedge:


adaptors/extensions.js
```diff
@@ -21,3 +21,35 @@ exports.clearTemporaryFile = function clearTemporaryFile(req, reply) {
 
   reply.continue();
 };
+
+exports.logRequest = function logRequest(req, reply) {
+  // We don't want to clutter the terminal, so only
+  // show request details if this was an error
+  if (req.response.isBoom) {
```
gideonthomas:
```js
if(!req.response.isBoom) {
  reply.continue();
  return;
}

var errorData = ...
...

reply.continue();
```

adaptors/extensions.js
```diff
@@ -21,3 +21,35 @@ exports.clearTemporaryFile = function clearTemporaryFile(req, reply) {
...
+    var errorData = req.response.data;
+    var errorObj = errorData && errorData.error;
+
+    // Prefer the error object stack over the
```
gideonthomas: +1

adaptors/extensions.js
```diff
@@ -21,3 +21,35 @@ exports.clearTemporaryFile = function clearTemporaryFile(req, reply) {
 
   reply.continue();
 };
+
+exports.logRequest = function logRequest(req, reply) {
+  // We don't want to clutter the terminal, so only
+  // show request details if this was an error
+  if (req.response.isBoom) {
+    var errorData = req.response.data;
```
gideonthomas: I would probably call this `response` and the next one `error`

adaptors/extensions.js
```diff
@@ -21,3 +21,35 @@ exports.clearTemporaryFile = function clearTemporaryFile(req, reply) {
...
+    var errorObj = errorData && errorData.error;
+
+    // Prefer the error object stack over the
+    // boom object stack
+    var stack = errorObj && errorObj.stack;
+    stack = stack ? stack : req.response.stack;
+
+    if (!req.response.data || req.response.data.debug) {
+      req.log.debug({
+        request: req,
```
gideonthomas: create this object outside since it does not change for debug vs. error

adaptors/extensions.js
```diff
@@ -21,3 +21,35 @@ exports.clearTemporaryFile = function clearTemporaryFile(req, reply) {
...
+        response: req.response,
+        error: req.response,
+        stack: stack
+      });
+    } else {
+      req.log.error({
+        request: req,
+        response: req.response,
+        error: req.response,
```
gideonthomas: just send the `errorObj` above for this. Makes the serializer logic easier.

adaptors/extensions.js
```diff
@@ -21,3 +21,35 @@ exports.clearTemporaryFile = function clearTemporaryFile(req, reply) {
...
+  if (req.response.isBoom) {
+    var errorData = req.response.data;
+    var errorObj = errorData && errorData.error;
+
+    // Prefer the error object stack over the
+    // boom object stack
+    var stack = errorObj && errorObj.stack;
+    stack = stack ? stack : req.response.stack;
+
+    if (!req.response.data || req.response.data.debug) {
```
gideonthomas:
```js
// Comment explaining why we're logging at debug level for !data
if(!data || data.debug) {
```

api/classes/errors.js
```diff
@@ -1,31 +1,43 @@
...
+  return reply(Boom.badRequest('`name` invalid', {
+    debug: true,
+    error: '`name` invalid'
+  }));
 };
 
 module.exports.generateErrorResponse = function(e) {
-  logger.error(e);
   if (e.isBoom) {
     return e;
```
gideonthomas: don't you still need to log this?

lib/logger.js
```diff
@@ -11,8 +11,44 @@ Hoek.assert(process.env.LOG_LEVEL, 'Must define LOG_LEVEL');
...
+}
+
+function responseSerializer(res) {
+  return {
+    payload: res.output.payload
+  };
+}
+
+function errorSerializer(res) {
+  if (!res.data) {
```
gideonthomas: In the log, just send `req.data.error` directly so that you don't have to do any of these checks

lib/logger.js
```diff
@@ -11,8 +11,44 @@ Hoek.assert(process.env.LOG_LEVEL, 'Must define LOG_LEVEL');
...
+    return;
+  }
+
+  if (typeof res.data.error === 'string') {
+    return {
+      message: res.data.error
+    };
+  }
+
+  var serializedData = JSON.parse(JSON.stringify(res.data.error));
```
gideonthomas: Let's try...catch to be safe

lib/logger.js
```diff
@@ -11,8 +11,44 @@ Hoek.assert(process.env.LOG_LEVEL, 'Must define LOG_LEVEL');
...
+  serializedData.message = res.data.error.message;
```
gideonthomas: This does not make sense to me...You clone the object above, so shouldn't it already have the `.message` property already attached?

api/classes/errors.js
sedge: This whole approach isn't ideal since it isn't clear how these pieces fit together.

But. This function ensures that whatever error is thrown in a prerequisite or handler function is wrapped as a `boom` by the time it hits the extension function in `extensions.js`. That's where the actual logging happens.

lib/logger.js
sedge: Remember how some attributes weren't being logged because they weren't enumerable? `res.data.error.message` is one of them, so I explicitly call it out here to make sure we see it.

lib/logger.js
sedge: I'm torn on this one. 

On the one hand, the entire point of the serializer is to parse this sort of stuff for us, allowing the user to log objects without worrying about what they contain. On the other hand, we have a dedicated function for doing the *only log in the entire request*, so it isn't unreasonable to do it there. 

I'm going to say these checks should stay here.

gideonthomas: The problem is that you're using bunyan for "abstract logging" (when usually it is used for logging different kinds of data and then selecting which set to use) which sorta negates the purpose of the serializer altogether. In fact, right now, you are doing two sets of serialization. I just think it makes more sense to do:
```js
{
  error: res.data.error
}
```
versus
```js
{
  error: res.data
}
```

But this is not something that is critical

lib/logger.js
gideonthomas: Can you add a comment above explaining this?

adaptors/extensions.js
sedge: No, it just moves that logic here since we don't want to log the error if `data` doesn't exist. I think it's cleaner to leave it as-is

lib/logger.js
sedge: If this throws, it won't crash the server. Instead, bunyan conveniently logs an error indicating that the serializer threw.

gideonthomas:
r=me

lib/logger.js
sedge: Doing our own try/catch doesn't make sense, since we'd only be replicating what bunyan does for us anyway: If there's an error, say so and keep going.

PR #164, opened 2015-09-22
Only upload/delete files from S3 that need to be uploaded/deleted

gideonthomas:
Instead of uploading every file to S3 when we hit "Publish", we decide which files to upload and only operate with them. Same case for unpublish.

### DIFF
--- a/api/classes/publisher.js
+++ b/api/classes/publisher.js
@@ -7,45 +7,47 @@ var client = require('../../lib/amazon-client').create();
 var log = require('../../lib/logger.js');
 var Remix = require('../../lib/remix');
 
-var File = require('../modules/files/model');
-var PublishedFiles = require('../modules/publishedFiles/model');
-var PublishedProject = require('../modules/publishedProjects/model');
+var Projects = require('../modules/projects/model');
+var PublishedProjects = require('../modules/publishedProjects/model');
+
+// SQL Query Generators
+var UserQueries = require('../modules/users/model').prototype.queries();
+var ProjectQueries = Projects.prototype.queries();
+var PublishedProjectQueries = PublishedProjects.prototype.queries();
+var PublishedFileQueries = require('../modules/publishedFiles/model').prototype.queries();
+
+var reject  = Promise.reject;
 
 /**
  * Utility functions
  */
-function reject(e) {
-  return Promise.reject(e);
+function success(type, username) {
+  return function(message) {
+    log.info('Publish for ' + username + ' - [' + type + '] ' + message);
+  };
 }
 
-function buildUrl(project, user) {
-  var url = process.env.PUBLIC_PROJECT_ENDPOINT + '/' +
-            user.get('name') + '/' +
-            project.get('id');
-  return url;
+function failure(type, username) {
+  return function(err) {
+    log.error({ error: err }, 'Publish for ' + username + ' - [' + type + ']');
+    reject(err);
+  };
 }
 
-/**
- * Model retrieval functions
- */
-function getUserForProject(project) {
-  return project.user().query({}).fetch();
+function getUploadRoot(user, project) {
+  if (!user || !project) {
+    return null;
+  }
+
+  return '/' + user.name + '/' + project.id;
 }
 
-function getPublishedProject(project) {
-  return PublishedProject.query({
-    where: {
-      id: project.get('published_id')
-    }
-  }).fetch();
-}
+function buildUrl(user, project) {
+  if (!user || !project) {
+    return null;
+  }
 
-function getPublishedFiles(publishedProject) {
-  return PublishedFiles.query({
-    where: {
-      published_id: publishedProject.get('id')
-    }
-  }).fetchAll();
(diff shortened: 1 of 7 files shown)


gideonthomas:
I cannot believe that this is actually a thing:
screen shot 2015-09-22 at 2 41 22 pm

gideonthomas:
Here is an example of what this does:
I did the following steps on Thimble:
* Create a new project and immediately publish
* Create a new file and then publish
* Update a file and then publish
* Delete a file and then publish
* Create, update and delete a file and then publish
* Unpublish the project

This is what the logs look like:
screen shot 2015-09-22 at 2 56 34 pm

api/classes/publisher.js
```diff
@@ -18,34 +23,33 @@ function reject(e) {
   return Promise.reject(e);
 }
```
sedge: This could probably become `var reject = Promise.reject`

api/classes/publisher.js
```diff
@@ -127,262 +120,276 @@ function deleteFileRemotely(file, root) {
...
-    if (files.length) {
-      return files.mapThen(function(file) {
-        return action(file, '/' + user.get('name') + '/' + project.get('id'), remixMetadata);
-      });
+
+/**
+ * Record fetching helpers
+ */
+function fetchProjectModel(id) {
+  return Projects.query({
```
sedge: Why aren't you doing something similar to the `UserQueries` approach below?

api/classes/publisher.js
```diff
@@ -127,262 +120,276 @@ function deleteFileRemotely(file, root) {
...
+  }).fetch();
+}
+
+function fetchUserForProject() {
+  var self = this;
+
+  return UserQueries.getOne(self.project.user_id)
+  .then(function(user) {
+    self.user = user;
+    return user;
```
sedge: Don't bother with this.

api/modules/publishedFiles/model.js
```diff
@@ -1,9 +1,101 @@
...
+          return publishedFiles[0];
+        });
+      },
+      getAllNewFiles: function(projectId) {
+        return new File().query()
+        .leftOuterJoin(self.tableName,
+                       FilesTable.column('id'),
+                       self.column('file_id'))
+        .where(FilesTable.column('project_id'), projectId)
+        .whereNull(self.column('file_id'))
```
sedge: Can you explain this line?

sedge: I'm surprised there's ever a time that a `publishedFile` has a null value for `file_id`

sedge: Or is this saying, "return the file record where no publishedFile references it"?

api/modules/publishedFiles/model.js
```diff
@@ -1,9 +1,101 @@
...
+                FilesTable.column('buffer'));
+      },
+      getAllModifiedFiles: function(publishedId) {
+        return new PublishedFile().query()
+        .innerJoin(FilesTable.tableName,
+                   self.column('file_id'),
+                   FilesTable.column('id'))
+        .where(self.column('published_id'), publishedId)
+        .where(function() {
+          this.whereRaw(self.column('path', null, true) +
```
sedge: This doesn't need to be returned?

api/modules/publishedFiles/model.js
```diff
@@ -1,9 +1,101 @@
...
+                        ' <> ' +
+                        FilesTable.column('path', null, true) +
+                        ' OR ' +
+                        self.column('buffer', null, true) +
+                        ' <> ' +
```
sedge: What does this operator mean?

sedge:
This is a massive improvement. A few questions, a couple of nits, but overall r+

api/modules/publishedFiles/model.js
gideonthomas: When you do a left outer join, it returns everything from the "left" table regardless of whether it refers to anything on the right table or not. We only want the values which do not refer to anything in the right table...so yeah...your second comment basically :P

api/modules/publishedFiles/model.js
gideonthomas: No coz you're setting it on `this`

api/modules/publishedFiles/model.js
gideonthomas: SQL for "not equal"

sedge:
Remove "successfully" from the upload/publish logs and put paths and project names in quotes

api/classes/publisher.js
gideonthomas: I'll add a comment above but this is the only method that does it since we accept a Bookshelf model in our entry point and we want to send back a Bookshelf model; this method allows us to do that. the `Queries` interface was done to return regular javascript objects since we tap directly into knex

api/classes/publisher.js
gideonthomas: Actually, it is used in the `createOrUpdatePublishedProject` method and I left it this way since it seemed to make more semantic sense - "fetch" should actually fetch which sorta implies return

Issue #163, opened 2015-09-22
extend the project definition with a "locked" property to prevent remixing


Pomax:
If Goggles needs to use the publish.wmo approach, then its projects should not be remixable, which requires passing an extra optional argument `editable:false` (or similar flag) into the project creation call. These projects should be published without the remix code added, and should reasonably not be loaded up by tools that can edit projects (which right now is just thimble).

sedge:
Why shouldn't goggles projects be remixable in thimble? Wouldn't that create a nice synergy between the tools if it was possible?

Pomax:
Because not everyone shares those ideals. Especially not when someone remixes license-encumbered content and we then host that content. Until legal tells us we can actually allow remixing for arbitrary web content, this will have to do.

Issue #159, opened 2015-09-22
projects should be able to a "which tool built this project" field


Pomax:
Right now publish.wmo assume the publishing tool is always Thimble, however with work underway on Goggles, this will not stay true for long. An additional field to indicate which publishing tool was used to build a project will be quite valuable for filtering/gallery purposes.

ashleygwilliams:
+1

Pomax:
Do we want this to be mandatory, so that other tools can't pretend to be Thimble by leaving off their client name, or should we just make this "if you fill in a tool name, we can filter on it, otherwise whatever"?

sedge:
I think "If you fill in a tool name we can filter it, otherwise whatever" is the smartest place to start.

Pomax:
agreed

PR #171, opened 2015-09-29
Schema retry

Pomax:
fixes https://github.com/mozilla/publish.webmaker.org/issues/159
fixes https://github.com/mozilla/publish.webmaker.org/issues/163

Adds a `readonly` and `client` column to the "projects" table, so that clients can indicate who they are, and tools can see what kind of operations might be allowed based on the readonly flag.

### DIFF
--- a/migrations/1_create_users_projects_files.js
+++ b/migrations/1_create_users_projects_files.js
@@ -1,84 +1,55 @@
-var Promise = require('bluebird');
+'use strict';
 
-exports.up = function (knex) {
-  return knex.schema
-  .createTable('publishedProjects', function(t) {
-    t.increments('id');
-    t.text('title').notNullable();
-    t.text('tags');
-    t.text('description');
+exports.up = function (knex, Promise) {
+  return Promise.resolve()
+  .then(function() {
+    return knex.schema.createTable('publishedProjects', function(t) {
+      t.increments('id');
+      t.text('title').notNullable();
+      t.text('tags');
+      t.text('description');
+    });
   })
-  .createTable('users', function (t) {
-    t.increments('id');
-    t.text('name').notNullable().unique();
+  .then(function() {
+    return knex.schema.createTable('users', function (t) {
+      t.increments('id');
+      t.text('name').notNullable().unique();
+    });
   })
-  .createTable('projects', function (t) {
-    t.increments('id');
-    t.integer('user_id').notNullable().references('id').inTable('users').onDelete('CASCADE');
-    t.integer('published_id').references('id').inTable('publishedProjects');
-    t.text('title').notNullable();
-    t.text('tags');
-    t.text('description');
-    t.text('publish_url');
-    t.text('date_created').notNullable();
-    t.text('date_updated').notNullable();
+  .then(function() {
+    return knex.schema.createTable('projects', function (t) {
+      t.increments('id');
+      t.integer('user_id').notNullable().references('id').inTable('users').onDelete('CASCADE');
+      t.integer('published_id').references('id').inTable('publishedProjects');
+      t.text('title').notNullable();
+      t.text('tags');
+      t.text('description');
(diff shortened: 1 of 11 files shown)


jbuck:
@Pomax can you split up the refactoring bits of code from the new feature bits of code into two separate PRs?

Pomax:
@jbuck I'd rather keep this all in one PR, as I had to put these refactors in in order for windows+psql+knex to work properly.

migrations/1_create_users_projects_files.js
```diff
@@ -1,84 +1,54 @@
...
+    return knex.schema.createTable('projects', function (t) {
+      t.increments('id');
+      t.integer('user_id').notNullable().references('id').inTable('users').onDelete('CASCADE');
+      t.integer('published_id').references('id').inTable('publishedProjects');
+      t.text('title').notNullable();
+      t.text('tags');
+      t.text('description');
+      t.text('publish_url');
+      t.text('date_created').notNullable();
+      t.text('date_updated').notNullable();
```
sedge: Add `readonly` and `client` here.

migrations/1_create_users_projects_files.js
```diff
@@ -1,84 +1,54 @@
...
+      t.text('path').notNullable();
+      t.specificType('buffer', 'bytea').notNullable();
     });
   });
 };
+
+
+exports.down = function (knex, Promise) {
+  // To prevent migrating down to "there aren't even any tables",
+  // there is no down() run.
```
sedge: wut?

migrations/2_alter_project_table.js
```diff
@@ -0,0 +1,18 @@
+exports.up = function (knex, Promise) {
+  return Promise.join(
+    knex.schema.table('projects', function (t) {
+      t.boolean('readonly');
```
sedge: Why is this an alter?

sedge: Clearly I missed something in this last week or so. It looks like we're trying to separate management of the database we're using for publish from the database we use to test publish.

If so, I'd prefer to see that land before this. Is there an issue on this (rather large) change to the workflow?

migrations/1_create_users_projects_files.js
Pomax: we *literally* deleted the production database because the initial migration is allowed to wipe the database. We're not doing that anymore.

migrations/2_alter_project_table.js
Pomax: no? This is an alter, because that's how migrations work. In order to extend the already in production database with two new columns that we need from now on, you need to alter the schema is a second migration.

migrations/1_create_users_projects_files.js
Pomax: absolutely not. This database schema is already used in production, any alterations to the table layout must be migrations.

migrations/2_alter_project_table.js
```diff
@@ -0,0 +1,18 @@
...
+      t.text('client');
+    })
+  );
+};
+
+exports.down = function (knex, Promise) {
+  // Note that this is irreversible, and will lead to data loss.
+  return Promise.join(
+    knex.schema.table('projects', function (t) {
+      t.dropColumn('readonly');
```
jbuck: I wonder if we should also have this do some reversible, like rename the column or something...

jbuck:
@Pomax Thanks for splitting up this patch.

I have one concern about the migration you're adding, where it has a destructive downgrade action. Could you consider changing that to a non-destructive downgrade so the next time I migrate the production database it doesn't get deleted?

r+ with nits

Pomax:
multi staging added:
1. create tables (irreversible)
2. create _readonly and _client (irreversible)
3. rename _readonly and _client to readonly and client (reversibly)

PR #172, opened 2015-09-29
Honour readonly

Pomax:
fixes https://github.com/mozilla/publish.webmaker.org/issues/173

builds on top of #171 for its schema extension/migration, the only new code on top of that PR here is https://github.com/Pomax/publish.webmaker.org/commit/149c4635f9f16f942abbd860e7a646634b27a70d (last commit in the series)

Turns off the injection of the "remix" script for `text/html` files if the project metadata contains the `readonly` flag (and it is set to `true`)

### DIFF
--- a/api/classes/publisher.js
+++ b/api/classes/publisher.js
@@ -100,7 +100,7 @@ function uploadFile(file, root, remixMetadata) {
   var path = root + file.get('path');
   var mimeType = mime.lookup(path);
 
-  if (mimeType === 'text/html') {
+  if (mimeType === 'text/html' && !remixMetadata.readonly) {
     buffer = new Buffer(Remix.inject(buffer.toString(), remixMetadata));
   }
 
@@ -188,7 +188,8 @@ exports.publish = function publish(project) {
         projectTitle: publishedProject.get('title'),
         projectAuthor: user.get('name'),
         dateUpdated: project.get('date_updated'),
-        host: Remix.resourceHost
+        host: Remix.resourceHost,
+        readonly: project.get('readonly')
       };
     });
   })


migrations/1_create_users_projects_files.js
```diff
@@ -1,84 +1,55 @@
...
+      t.integer('file_id').references('id').inTable('files');
+      t.text('path').notNullable();
+      t.specificType('buffer', 'bytea').notNullable();
     });
   });
 };
+
+
+exports.down = function (knex) {
+  // We cannot revert beyond "having tables".
```
gideonthomas: Can the reversion be "remove all data from the tables"

seeds/3_files.js
```diff
@@ -3,46 +3,48 @@
...
 exports.seed = function(knex, Promise) {
+  // As knex does not have an .insertIfNotExists(), we hook into catch()
+  // so that seeding doesn't break when running on an already seeded database.
   return Promise.join(
     knex('files').insert({
       project_id: 1,
       path: '/spacecats-API/index.js',
       buffer: fs.readFileSync(__dirname + '/file_data/index.js')
-    }),
+    }).catch(function() { return knex('files'); }),
```
gideonthomas: I think you should be able to attach just one `catch` at the end of the `Promise.join` call since they all do the same thing.

migrations/1_create_users_projects_files.js
Pomax: nope. We had that. Then we accidentally wiped the thimble production database. It is too dangerous to leave in, and doing testing better will be tackled in https://github.com/mozilla/publish.webmaker.org/issues/174

seeds/3_files.js
Pomax: this has been removed again based on the conclusion that the seeds are for testing only, so will always operate on an empty db, where there is no risk over appending to preexisting data (this PR builds on top of https://github.com/mozilla/publish.webmaker.org/pull/171 which got updated to reflect that)

sedge: Seeding the DB used by the service is valuable in development. What's stopping us from having a solution that allows both?

Pomax: that solution is https://github.com/mozilla/publish.webmaker.org/issues/174#issuecomment-144783185

PR #177, opened 2015-10-06
Distribute tar file creation across multiple turns of the event loop

humphd:
The code that creates a tarball will sometimes timeout the server, since a big project with lots of (possibly big) files will all need to get processed.  This fix is a first-level attempt to fix that, by splitting the work across multiple turns of the event loop.  By doing so, bits of the work will happen, but not such that it blocks the whole process.  A more elaborate fix would do it in a secondary process, but I want to see if we can improve things with something simpler.

Based on my reading, I think I want `setImmediate` vs. `process.nextTick`, but feel free to tell me I'm wrong if you can back it up!

### DIFF
--- a/api/classes/base_controller.js
+++ b/api/classes/base_controller.js
@@ -87,27 +87,32 @@ BaseController.prototype.getAllAsMeta = function(req, reply) {
   reply(req.generateResponse(req.pre.records.toJSON()));
 };
 
+// NOTE: creating the tarball for a project can be a lengthy process, so
+// we do it in multiple turns of the event loop, so that other requests
+// don't get blocked.
 BaseController.prototype.getAllAsTar = function(req, reply) {
   var files = req.pre.records.models;
   var tarStream = Tar.pack();
-  var self = this;
+  var model = this.Model;
 
-  function startStreaming(files, model) {
-    return Promise.map(files, function(file) {
-      return model.query({
-        where: {
-          id: file.get('id')
-        },
-        columns: ['buffer']
-      }).fetch().then(function(model) {
+  function processFile(file) {
+    return model.query({
+      where: {
+        id: file.get('id')
+      },
+      columns: ['buffer']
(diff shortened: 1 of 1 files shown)


humphd:
@jbuck can you r? this for me, too?  Make sure I'm clear on this.

gideonthomas:
I think this looks right!

api/classes/base_controller.js
```diff
@@ -87,27 +87,32 @@ BaseController.prototype.getAllAsMeta = function(req, reply) {
...
         tarStream.entry({ name: file.get('path') }, model.get('buffer'));
       });
-    }).then(function() {
-      tarStream.finalize();
     });
   }
 
-  startStreaming(files, self.Model).catch(errors.generateErrorResponse);
+  setImmediate(function() {
+    Promise.map(files, processFile)
```
jbuck: I wonder if you'd also want to specify a [concurrency option](https://github.com/petkaantonov/bluebird/blob/master/API.md#option-concurrency) here, to limit the number of in-flight database queries

jbuck:
@humphd only one possible change here. Let me know what you think of that!

api/classes/base_controller.js
humphd: I think this sounds like a good idea.  Do you have a suggestion as to what I should set it to?  The docs don't really give much help here, just indicate that you can limit it.  What's a good number, do you think?

jbuck: It looks like the default connection pooling setup for bookshelf is 10 maximum connections, so we definitely want it lower than that. Ball-parking it... maybe limit it to 2 simultaneous entries at once?

humphd:
@jbuck, updated.  You cool with this?

jbuck:
@humphd r+ with nits around the style that travis is complaining about

humphd:
Cool, thanks you two.  We should ship this to staging/prod, too.

PR #184, opened 2015-10-13
Do not serialize buffers for response on POST/PUT requests for files

gideonthomas:


api/modules/files/controller.js
```diff
@@ -8,6 +8,11 @@ var controller = new BaseController(Model);
 
 var PublishedFiles = require('../publishedFiles/model');
 
+function formatResponse(data) {
```
humphd: I wonder if there is a better name for this, which indicates that we're intentionally stripping the file contents, and only returning the response metadata?  At the very least, I'd put a comment above this to explain the thinking.

humphd:
One nit, r+ with that.  Let's test on staging.

PR #185, opened 2015-10-20
Fix mozilla/thimble.webmaker.org#1184 - Do not inject remix code for readonly projects

gideonthomas:


gideonthomas:
@Pomax nice catch :)
Could you r when you get a chance?

api/classes/publisher.js
```diff
@@ -73,7 +73,7 @@ function uriSafe(path) {
  */
 function upload(path, buffer, remixMetadata) {
   var mimeType = mime.lookup(path);
-  if (mimeType === 'text/html') {
+  if (mimeType === 'text/html' && remixMetadata) {
```
Pomax: it's better to rely on `remixMetaData.readonly` here (1/2)

api/classes/publisher.js
```diff
@@ -178,6 +178,10 @@ function updateProjectDetails() {
 function setRemixDataForPublishedProject() {
   var self = this;
 
+  if (self.project.readonly) {
```
Pomax: (2/2) and then not do this, but build the real meta data object, and assign the `readonly` flag as `true`.

gideonthomas:
@Pomax fixed! Does it look good now?

Pomax:
r+

PR #186, opened 2015-10-26
Fix mozilla/thimble.webmaker.org#1170 - Track date created and updated for published projects

gideonthomas:


gideonthomas:
@Pomax can you please r when you get a chance?

api/classes/publisher.js
```diff
@@ -194,7 +194,8 @@ function createOrUpdatePublishedProject() {
   var projectData = {
     title: project.title,
     tags: project.tags,
-    description: project.description
+    description: project.description,
+    _date_updated: (new Date()).toISOString()
```
Pomax: why the underscore? just use `date_updated`

api/classes/publisher.js
```diff
@@ -202,6 +203,7 @@ function createOrUpdatePublishedProject() {
     if (publishedProject) {
       return PublishedProjectQueries.updateOne(publishedProject.id, projectData);
     } else {
+      projectData._date_created = projectData._date_updated;
```
Pomax: no underscores - these are real columns.

migrations/4_add_date_tracking_published_projects.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
```
Pomax: if this is data that must not be lost, you migrate in two steps:
1. in the first migration, add the columns as _colname, with a down migration that does nothing, and then
2. in the second migration, rename the columns to their `colname` rather than `_colname`, with a down migration that simply reverses that rename.

Pomax:
Have a look at https://github.com/mozilla/publish.webmaker.org/pull/171/files as well, which was the update for adding the `readonly` and `client` fields, which shows the two-step migration.

You'll also want to add those changes to the tests now, I would imagine.

gideonthomas:
Yeah...there aren't really any tests for the publishedProjects and publishedFiles modules :P

migrations/5_toggle_date_tracking_published_projects.js
```diff
@@ -0,0 +1,21 @@
...
+  return Promise.join(
+    knex.schema.table('publishedProjects', function (t) {
+      t.renameColumn('_date_created', 'date_created');
+      t.renameColumn('_date_updated', 'date_updated');
+    })
+  );
+};
+
+exports.down = function (knex, Promise) {
+  // "Disable" the readonly and client columns.
```
Pomax: y'arr that are not theyr names.

Pomax:
I possible, it would be great to, at the very least, have schema testing (a la projects/create.js and files/create.js) to make ensure that the migration even worked (we know it should, but computers are dumb, and sometimes they do the wrong thing).

If we extend https://github.com/mozilla/publish.webmaker.org/tree/master/test/lib/fixtures to have the same subdirs as https://github.com/Pomax/publish.webmaker.org/tree/ea4db88bbd55091813d44678070f90cd717ac436/api/modules, then we can add a seed for publishedProjects (for now, since we're not fiddling with publishFiles) that drops some data in there, and then add a file like https://github.com/Pomax/publish.webmaker.org/blob/ea4db88bbd55091813d44678070f90cd717ac436/test/api/handlers/files/create.js that tests the schema and associated project.

The "logical" time to do this would be with this PR, but if that holds up other work, then this should be a followup to be done "within the week" or something, to make sure it doesn't get ignored and then lost to time =)

gideonthomas:
@Pomax do the tests look ok?

seeds/4_publishedProjects.js
```diff
@@ -0,0 +1,15 @@
+exports.seed = function(knex, Promise) {
+  return Promise.join(
+    knex('publishedProjects').insert({
+      title: 'sinatra-contrib',
+      tags: 'ruby, sinatra, community, utilities',
+      description: 'Hydrogen atoms Sea of Tranquility are creatures of the cosmos shores of the cosmic ocean.',
+      date_created: '2015-06-03T16:21:58+00:00',
```
Pomax: This does not look like a proper simplified ISO time, which would be `2015-06-03T16:21:58.000Z'

test/api/handlers/publishedProjects/get-one.js
```diff
@@ -0,0 +1,74 @@
...
+
+    server.inject(options, function(response) {
+      expect(response.statusCode).to.equal(200);
+
+      expect(response.result).to.exist();
+      expect(response.result.id).to.be.a.number();
+      expect(response.result.title).to.be.a.string();
+      expect(response.result.tags).to.be.a.string();
+      expect(response.result.description).to.be.a.string();
+      expect(response.result.date_created).to.be.a.string();
```
Pomax: having dates be strings is still very worrisome. We may want to fix this sooner than later =(

PR #192, opened 2015-11-03
Update dates for existing published projects

gideonthomas:
This patch does two things:

1. It fixes a seeding issue caused by the improper use of Promises which lead to an attempt to update the `projects` table with a `published_id` which references a published project in the `publishedProjects` table that hadn't been created yet.
2. Removes null values for the `date_created` and `date_updated` fields in the `publishedProjects` table caused by published projects that existed before the previous migration was landed. 
 * For those published projects that were re-published after the date tracking feature was added, they only miss a `date_created`. So, we set it to the `date_updated` value to assume that it was created for the first time when it was updated. 
 * For those published projects that were not re-published after the date tracking feature was added, they miss the `date_created` and `date_updated` fields. So, we set them to the `date_updated` for their corresponding project in the `projects` table.

### DIFF
--- a/migrations/6_update_dates_existing_published_projects.js
+++ b/migrations/6_update_dates_existing_published_projects.js
@@ -1,0 +1,45 @@
+'use strict';
+
+exports.up = function (knex, Promise) {
+  return knex('publishedProjects')
+  .whereNull('date_created')
+  .orWhereNull('date_updated')
+  .select('id', 'date_created', 'date_updated')
+  .then(function(publishedProjects) {
+    return Promise.map(publishedProjects, function(publishedProject) {
+      var publishedProjectId = publishedProject.id;
+      var dateUpdated = publishedProject.date_updated;
+
+      // For existing published projects that have been updated since
+      // date tracking has been added, they will have a date_updated
+      // but not a date_created. Set the date_created to the date_updated
+      if (dateUpdated) {
+        return knex('publishedProjects').update('date_created', dateUpdated)
+        .where('id', publishedProjectId);
+      }
+
+      // For existing published projects that have not been updated
+      // since date tracking has been added - set both the dates
+      // to the date_updated of the corresponding project
+      return knex('projects')
+      .where('published_id', publishedProjectId).select('date_updated')
+      .then(function(projects) {
+        var date = projects[0].date_updated;
+
+        // Update the date_created and date_updated fields in the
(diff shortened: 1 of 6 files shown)


gideonthomas:
@Pomax would you be able to review this?
Also, can you try to pull my branch down and run the tests? Somehow the tests are failing for me locally (for code that I have not even touched) but they pass here on Travis

Pomax:
I might be able to, but after I get to London.

gideonthomas:
That's cool, this doesn't need to be done urgently so even after MozFest is fine. Basically whenever you get some free time :P

migrations/6_update_dates_existing_published_projects.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+
+exports.up = function (knex, Promise) {
+  return knex('publishedProjects').select('id', 'date_created', 'date_updated')
```
jbuck: You should add a `WHERE` clause here to filter out projects that already have a date_created and date_updated value. Something like `WHERE date_created IS NOT NULL AND date_updated IS NOT NULL`

migrations/6_update_dates_existing_published_projects.js
```diff
@@ -0,0 +1,48 @@
...
+
+      // For existing published projects that have not been updated
+      // since date tracking has been added - set both the dates
+      // to the date_updated of the corresponding project
+      return knex('projects')
+      .where('published_id', publishedProjectId).select('date_updated')
+      .then(function(projects) {
+        var date = projects[0].date_updated;
+
+        return Promise.join(
```
jbuck: Can you combine these two queries together somehow? I don't know the knex API well enough, but in SQL it'd be something like `UPDATE publishedProjects SET date_created = $1, date_updated = $2 WHERE id = $3`

jbuck:
@gideonthomas I added some suggestions

Pomax:
I've not had time to review yet, but where will we be implementing date control from creation vs update on the publish side, so that the format and values are centrally controlled, rather than (potentially incorrectly) specified by each client? If we're not doing that in this PR, we'll want to make sure to do that as fast follow-up

gideonthomas:
@Pomax np at all! I have that in an incoming pr actually! should be done by this week. Will ping you in it for review.

jbuck:
lgtm, r+

PR #195, opened 2015-11-09
Use datetime sql type to store project dates

gideonthomas:
This builds on https://github.com/mozilla/publish.webmaker.org/pull/192 and relies on it landing first.

This PR changes the existing date columns stored as text in the `projects` and `publishedProjects` tables to use the datetime sql type instead.


gideonthomas:
Depends on https://github.com/mozilla/publish.webmaker.org/pull/200 landing first

PR #197, opened 2015-11-25
Make sure that the select queries return records otherwise, bail

gideonthomas:


migrations/6_update_dates_existing_published_projects.js
```diff
@@ -6,6 +6,10 @@ exports.up = function (knex, Promise) {
   .orWhereNull('date_updated')
   .select('id', 'date_created', 'date_updated')
   .then(function(publishedProjects) {
+    if(!publishedProjects) {
+      return;
```
jbuck: Maybe `return Promise.resolve();`? I'm not sure if that's needed or not

jbuck:
r+

PR #203, opened 2015-12-07
Issue165

gideonthomas:
The values for `date_created` and `date_updated` for `projects` as well as `publishedProjects` will now be generated on the server and do not need to be passed in by the client. 
This is also true for the `/remix` route which took the date in the query string. 
(Dates will still be accepted in the request payload so that existing services that depend on publish do not break, but they will not be used).

### DIFF
--- a/api/classes/publisher.js
+++ b/api/classes/publisher.js
@@ -195,7 +195,7 @@ function createOrUpdatePublishedProject() {
     title: project.title,
     tags: project.tags,
     description: project.description,
-    date_updated: new Date()
+    date_updated: (new Date()).toISOString()
   };
 
   return fetchPublishedProject.call(self)
--- a/api/classes/publisher.js
+++ b/api/classes/publisher.js
@@ -182,7 +182,7 @@ function setRemixDataForPublishedProject() {
     projectId: self.publishedProject.id,
     projectTitle: self.publishedProject.title,
     projectAuthor: self.user.name,
-    dateUpdated: self.project.date_updated,
+    dateUpdated: self.publishedProject.date_updated.toISOString(),
     host: Remix.resourceHost,
     readonly: self.project.readonly
   };
--- a/api/modules/projects/controller.js
+++ b/api/modules/projects/controller.js
@@ -31,19 +31,24 @@ function formatResponse(model) {
 }
 
 controller.formatRequestData = function(req) {
+  var now = new Date();
   var data = {
     title: req.payload.title,
     user_id: req.payload.user_id,
     tags: req.payload.tags,
     description: req.payload.description,
-    date_created: req.payload.date_created,
-    date_updated: req.payload.date_updated,
+    date_created: now,
+    date_updated: now,
     readonly: req.payload.readonly,
     client: req.payload.client
   };
+
+  // If it is an update request
   if (req.params.id) {
     data.id = parseInt(req.params.id);
+    delete data.date_created;
   }
+
   return data;
 };
 
--- a/api/modules/projects/controller.js
+++ b/api/modules/projects/controller.js
@@ -11,24 +11,7 @@ var PublishedFiles = require('../publishedFiles/model');
 var Model = require('./model');
 var controller = new BaseController(Model);
 
-// Taken from http://stackoverflow.com/a/643827
-function isDate(val) {
-  return Object.prototype.toString.call(val) === '[object Date]';
-}
-
-function formatResponse(model) {
-  var created = model.get('date_created');
-  var updated = model.get('date_updated');
-
-  if (isDate(created)) {
-    model.set('date_created', created.toISOString());
-  }
-  if (isDate(updated)) {
-    model.set('date_updated', updated.toISOString());
-  }
-
-  return model;
-}
+var dateTracker = require('../../../lib/utils').dateTracker;
 
 controller.formatRequestData = function(req) {
   var now = new Date();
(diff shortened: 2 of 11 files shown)


gideonthomas:
This patch depends on https://github.com/mozilla/publish.webmaker.org/pull/195 landing first.

api/classes/publisher.js
```diff
@@ -195,7 +195,7 @@ function createOrUpdatePublishedProject() {
     title: project.title,
     tags: project.tags,
     description: project.description,
-    date_updated: (new Date()).toISOString()
+    date_updated: new Date()
```
Pomax: this will use `toString()`, which is not necessarily the same as `toISOString()` - better to keep it explicitly ISO?

api/modules/projects/controller.js
```diff
@@ -11,28 +11,72 @@ var PublishedFiles = require('../publishedFiles/model');
 var Model = require('./model');
 var controller = new BaseController(Model);
 
+// Taken from http://stackoverflow.com/a/643827
+function isDate(val) {
+  return Object.prototype.toString.call(val) === '[object Date]';
```
Pomax: neat, but in this case I'd say `instanceof` makes sense:
```
return val instanceof Date;
```

api/modules/projects/controller.js
```diff
@@ -11,28 +11,72 @@ var PublishedFiles = require('../publishedFiles/model');
...
+  }
+  if (isDate(updated)) {
+    model.set('date_updated', updated.toISOString());
+  }
+
+  return model;
+}
+
 controller.formatRequestData = function(req) {
+  var now = new Date();
```
Pomax: `var now = Date.now()` ?

api/modules/projects/schema.js
```diff
@@ -3,8 +3,8 @@ var Joi = require('joi');
 module.exports = Joi.object().keys({
   title: Joi.string().required(),
   user_id: Joi.number().integer().required(),
-  date_created: Joi.string().required(),
-  date_updated: Joi.string().required(),
+  date_created: Joi.date(),
+  date_updated: Joi.date(),
```
Pomax: :heart:

api/modules/publishedProjects/controller.js
```diff
@@ -15,6 +15,44 @@ function ensureRemixSuffix(title) {
   return title.replace(/( \(remix\))*$/, ' (remix)');
 }
 
+// Taken from http://stackoverflow.com/a/643827
+function isDate(val) {
```
Pomax: code duplication from  api/modules/projects/controller.js, so this should live in a util file that is required by both controllers.

api/modules/publishedProjects/controller.js
```diff
@@ -15,6 +15,44 @@ function ensureRemixSuffix(title) {
...
+  return Object.prototype.toString.call(val) === '[object Date]';
+}
+
+function formatResponse(model) {
```
Pomax: idem ditto; code duplication

api/modules/publishedProjects/controller.js
```diff
@@ -46,15 +84,18 @@ controller.remix = function(req, reply) {
   }
 
   function duplicateProject() {
+    var now = new Date();
```
Pomax: `(new Date()).toISOString()`?

api/modules/publishedProjects/model.js
```diff
@@ -13,6 +13,34 @@ var instanceProps = {
   publishedFiles: function() {
     return this.hasMany(require('../publishedFiles/model'));
   },
+  format: function(model) {
```
Pomax: code duplication; it makes more sense to refactor this to its own file and using something like
```
format: require('utils/datehandling/format'),
pares: require('utils/datehandling/parse'),
...
```

api/modules/publishedProjects/model.js
```diff
@@ -13,6 +13,34 @@ var instanceProps = {
...
+      }
+      if (model.date_updated) {
+        model._date_updated = model.date_updated;
+        delete model.date_updated;
+      }
+    }
+
+    return model;
+  },
+  parse: function(model) {
```
Pomax: see above note

migrations/20151117173053_add_datetime_columns.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
```
Pomax: Making the timestamps wider than 1 number is a good idea, but instead of 20151117173053, let's just use 0001, 0002, etc.

seeds/4_publishedProjects.js
```diff
@@ -4,8 +4,8 @@ exports.seed = function(knex, Promise) {
       title: 'sinatra-contrib',
       tags: 'ruby, sinatra, community, utilities',
       description: 'Hydrogen atoms Sea of Tranquility are creatures of the cosmos shores of the cosmic ocean.',
-      date_created: '2015-06-19T17:21:58.000Z',
-      date_updated: '2015-06-23T06:41:58.000Z'
+      _date_created: new Date('2015-06-19T17:21:58.000Z'),
```
Pomax: this format is different from the formats uesd in `seeds/2_projects.js`.

test/api/handlers/projects/create.js
```diff
@@ -27,6 +27,31 @@ after(function(done) {
   server.stop(done);
 });
 
+function cleanup(id, callback) {
+  server.inject({
+    url: '/projects/' + id,
+    method: 'delete',
+    headers: {
+      authorization: 'token ag-dubs'
```
Pomax: can we change this to a `test` token, so we're not baking people's usernames into our tests?

migrations/20151117173053_add_datetime_columns.js
gideonthomas: I think the timestamp might be better because knex generates a migration template for you with one command and it timestamps it based on this format. I feel like for consistency, following whatever knex does would be easier

test/api/handlers/projects/create.js
gideonthomas: I can, but would prefer to do it in a follow-up patch

api/modules/projects/controller.js
gideonthomas: not in this case. This method is called before any server-side processing happens. And I think you would want to work with date objects when you process the request. I'll change the database level interaction that happens so that just before it communicates with the db, it converts the dates to ISO strings

seeds/4_publishedProjects.js
gideonthomas: I changed the one in `seeds/2_projects.js`

gideonthomas:
@Pomax re-r?

migrations/20151117173053_add_datetime_columns.js
Pomax: ah I see. aiight let's keep it this then!

test/api/handlers/projects/create.js
Pomax: that works for me. Want to file that?

Pomax:
looks pretty solid, npm tests / knex all look like they pass. I'd say R+

PR #204, opened 2015-12-14
Fix 685f582 to actually not serialize buffers

gideonthomas:


humphd:
r=me

humphd:
I really feel like this had a huge positive effect, esp on Memory usage:

screenshot 2015-12-15 22 14 41

@gideonthomas++

PR #208, opened 2015-12-16
#4 - Use only the dateTime columns

gideonthomas:


cadecairos:
R+

gideonthomas:
Note to self: before landing this, the last commit which involved a migration in it should be run on production. The last attempt to do so was unsuccessful due to exceeded memory and hence should be completed successfully before landing this commit.

PR #211, opened 2015-12-18
Fix date copying migration to deal with number strings

gideonthomas:


migrations/20151120164525_copy_text_date_columns_into_datetime_columns.js
```diff
@@ -1,5 +1,10 @@
 'use strict';
 
+function resolveType(dateStr) {
+  var num = parseInt(dateStr);
```
cadecairos: lets pass in a radix param, so that JS doesn't make assumptions for us.

cadecairos:
R+ with one small addition

PR #214, opened 2016-01-12
Fix #189 - Upload content with UTF-8 encoding to AWS

### LINKED ISSUE #189: make sure we upload files to S3 with utf8 encoding
right now they seem to be uploaded without explicit encoding, which can throw off the text encoding detection in the browser.

gideonthomas:
@Pomax r please?

### DIFF
--- a/api/classes/publisher.js
+++ b/api/classes/publisher.js
@@ -78,7 +78,7 @@ function upload(path, buffer, remixMetadata) {
   }
   var headers = {
     'Cache-Control': 'max-age=0',
-    'Content-Type': mimeType,
+    'Content-Type': mimeType + '; charset=utf-8',
     'Content-Length': buffer.length
   };
 


api/classes/publisher.js
```diff
@@ -78,7 +78,7 @@ function upload(path, buffer, remixMetadata) {
   }
   var headers = {
     'Cache-Control': 'max-age=0',
-    'Content-Type': mimeType,
+    'Content-Type': mimeType + '; charset=utf-8',
```
Pomax: Small nit: officially it's `UTF-8`, rather than `utf-8`. Case doesn't always matter, so better to be safe.

Pomax: w00p, never mind, http://tools.ietf.org/html/rfc2616#section-3.4 tells me it's case insensitive =P

Pomax:
R+

PR #217, opened 2016-01-13
Remove _date_created if it is null

gideonthomas:
@cadecairos r? 
Unfortunately, I had to account for the null case as well and forgot :(


gideonthomas:
Actually, this still needs some work

cadecairos:
tests passing for me, travis is green, R+

PR #223, opened 2016-03-08
Add route to update file path

gideonthomas:
Needed for: https://github.com/mozilla/thimble.mozilla.org/issues/1311

### DIFF
--- a/adaptors/extensions.js
+++ b/adaptors/extensions.js
@@ -1,4 +1,6 @@
-var fs = require('fs');
+'use strict';
+
+var fs = require(`fs`);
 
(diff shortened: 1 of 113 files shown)


gideonthomas:
@Pomax when you review this, you can ignore the second commit. Apparently the old jshint/jscs files don't exist in `mofo-style` anymore (and ay previous tags containing them no longer exist) so I had to add eslint and experience an awful nightmare of me fixing a billion errors (even after `--fix`) so that travis would build.

gideonthomas:
@Pomax sorry, I would hold off on the review until later on (I will ping you). I apparently need to add some more logic here to handle issues with the publishing workflow

Issue #224, opened 2016-04-26
publish to /http/ prefixed S3 location if client_id is x-ray goggles.


Pomax:
Addresses https://github.com/MozillaFoundation/mofo-devops/issues/324, this issue is for making publish.wmo publish to S3 with an `/HTTP/` (uppercase) prefix when the project's client_id is x-ray goggles' client id.

This will allow us to set up special routing that lets people remix http:// websites without it necessarily ending up seeing a failed project due to mixed content.

(Note that the prefix is not `http` as this might conflict with a user called "http")

PR #225, opened 2016-04-26
added HTTP namespacing for whitelisted clients

Pomax:
fixes https://github.com/mozilla/publish.webmaker.org/issues/224 by updating `buildURL` and `getUploadRoot` with a `client` argument, so we can adaptively set the upload root with or without an `/HTTP/` prefix depending on whether the client matches a whitelist for http publication.

### DIFF
--- a/api/classes/publisher.js
+++ b/api/classes/publisher.js
@@ -34,20 +34,30 @@ function failure(type, username) {
   };
 }
 
-function getUploadRoot(user, project) {
+function getUploadRoot(client, user, project) {
   if (!user || !project) {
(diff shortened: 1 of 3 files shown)


gideonthomas:
r+

gideonthomas:
@Pomax would you be able to update this:
https://github.com/mozilla/publish.webmaker.org/blob/master/package.json#L58
to 1.0.0 or something so that travis passes?

Pomax:
done and done.

gideonthomas:
r+

PR #228, opened 2016-06-02
Allow exporting a project as a tarball

gideonthomas:


xmatthewx:
Excited to see export as a feature in tools. Will this be added to the Thimble frontend? Super useful. Question: will most people need a utility to unpack it? Or is that now standard on windows and mac? If so, awesome. If not, I'd consider changing this to Zip. (Forgive this drive by comment if my question has totally missed the mark on the purpose of this.)

gideonthomas:
@xmatthewx this pr is mainly relevant for debugging purposes. Often we find ourselves struggling to access someone's project when they report something like "I can't access my project" or "my project isn't saving". Even if we do get the project, we end up having to do a bunch of manual steps to get that project into Thimble so that we can try to reproduce the error that the user sees. This will export their project (if we know their project id) and we can just drag and drop it into Thimble to reproduce the error, hence cutting down all those manual steps.

This was totally my bad. I should have added a description to this patch to clarify its purpose, so sorry for the confusion.

The export feature that you are referring to already exists in Thimble (and it exports as .zip 👍 ). Here's how you can try it out: https://github.com/mozilla/thimble.mozilla.org/wiki/Using-Thimble#export-a-project

xmatthewx:
Excellent. Thanks for all the info. Sorry for the distraction!

cadecairos:
lgtm!

PR #230, opened 2016-06-24
ES6 upgrade

gideonthomas:
**[PLEASE READ]**

This patch converts *most* of publish to ES6 syntax to make it easier to keep up with updates to the JS standards as well as make the code cleaner and slightly easier to read.

We follow the mofo-style ESLinting rules and conventions with a few overrides as needed for this project (which can be seen in the `.eslintrc.yaml` file).

For the sake of sanity, I have tried to keep the commits as distinct as possible with a logical ordering of changes based on MVC. 

*Note to reviewer - it might be easier to review in the order of the commits. For the review, I'm mainly looking for comments on clear logic errors or performance hits that might be a result of this code change, rather than stylistic comments (if you have any, feel free to add "nits").*

### DIFF
--- a/adaptors/extensions.js
+++ b/adaptors/extensions.js
@@ -1,20 +1,22 @@
-var fs = require('fs');
+"use strict";
 
-exports.clearTemporaryFile = function clearTemporaryFile(req, reply) {
-  if (req.response.isBoom) {
+const fs = require(`fs`);
+
+exports.clearTemporaryFile = function(request, reply) {
+  if (request.response.isBoom) {
     return reply.continue();
   }
 
   // Once a successful request completes, we delete any
   // temporary files we created
-  req.response.once('finish', function() {
-    if (!req.app.tmpFile) {
+  request.response.once(`finish`, function() {
+    if (!request.app.tmpFile) {
       return;
     }
 
-    fs.unlink(req.app.tmpFile, function(err) {
+    fs.unlink(request.app.tmpFile, function(err) {
       if (err) {
-        req.log.error('Failed to destroy temporary file with ' + err);
+        request.log.error(`Failed to destroy temporary file with ${err}`);
       }
     });
   });
@@ -22,38 +24,40 @@ exports.clearTemporaryFile = function clearTemporaryFile(req, reply) {
   reply.continue();
 };
 
-exports.logRequest = function logRequest(req, reply) {
+exports.logRequest = function(request, reply) {
+  const response = request.response;
+
   // We don't want to clutter the terminal, so only
   // show request details if this was an error
-  if (!req.response.isBoom) {
+  if (!response.isBoom) {
     reply.continue();
     return;
   }
 
-  var data = req.response.data;
-  var error = data && data.error;
+  const data = response.data;
+  const error = data && data.error;
 
   // Prefer the error object stack over the
   // boom object stack
-  var stack = error && error.stack || req.response.stack;
+  const stack = error && error.stack || response.stack;
 
-  var logLevel = 'error';
+  let logLevel = `error`;
+
   if (!data || data.debug) {
     // Errors we process will contain a "data" property
     // containing the error object (or string) and the
     // level of the error. If it doesn't exist, then the `boom`
     // object was created by the framework and represents an
     // error we don't care about under normal circumstances
-    logLevel = 'debug';
+    logLevel = `debug`;
   }
 
-  req.log[logLevel]({
-    request: req,
-    response: req.response,
-    error: error,
-    stack: stack
+  request.log[logLevel]({
+    request,
+    response,
+    error,
+    stack
   });
 
   reply.continue();
 };
-
--- a/adaptors/plugins.js
+++ b/adaptors/plugins.js
@@ -1,4 +1,6 @@
+"use strict";
+
 module.exports = [
-  require('lout'),
-  require('hapi-auth-bearer-token')
+  require(`lout`),
+  require(`hapi-auth-bearer-token`)
 ];
--- a/api/classes/base_controller.js
+++ b/api/classes/base_controller.js
@@ -1,7 +1,9 @@
-var Boom = require('boom');
-var Promise = require('bluebird'); // jshint ignore:line
-var Tar = require('tar-stream');
-var errors = require('./errors');
+"use strict";
+
+const Boom = require(`boom`);
+const Promise = require(`bluebird`);
+const Tar = require(`tar-stream`);
+const Errors = require(`./errors`);
 
 // This method is used to format the response data for
 // HTTP methods that modify data (e.g. POST, PUT, etc.)
@@ -11,133 +13,170 @@ function defaultFormatResponse(model) {
   return model;
 }
 
-function BaseController(model)  {
-  this.Model = model;
-}
+class BaseController {
+  constructor(Model) {
+    this.Model = Model;
+  }
 
-BaseController.prototype.formatRequestData = function(req) {
-  // Abstract base method,
-  // formats data for database entry
-};
+  /* eslint no-unused-vars: ["error", { "argsIgnorePattern": "^request$" }]*/
+  formatRequestData(request) {
+    // Abstract base method; formats data for database entry
+    throw new Error(`formatRequestData has not been implemented in subclass`);
+  }
 
-BaseController.prototype.formatResponseData = function(model) {
-  // Base method to format data that will be sent in the response
-  // By default, it returns the model as is
-  return model;
-};
+  formatResponseData(model) {
+    // Base method to format data that will be sent in the response
+    // By default, it returns the model as is
+    return model;
+  }
 
-BaseController.prototype.getOne = function(req, reply) {
-  var record = this.formatResponseData(req.pre.records.models[0].toJSON());
-  reply(req.generateResponse(record)
-    .code(200));
-};
+  // `formatResponse` is an optional processing function that can be passed
+  // in to modify what is sent in the response body. If no function is
+  // provided, the full model for the current method is used in the response.
+  create(request, reply, formatResponse) {
+    const requestData = this.formatRequestData(request);
 
-BaseController.prototype.getAll = function(req, reply) {
-  var records = req.pre.records.toJSON().map(this.formatResponseData);
-  reply(req.generateResponse(records));
-};
-
-// `formatResponse` is an optional processing function that can be passed
-// in to modify what is sent in the response body. If no function is
-// provided, the full model for the current method is used in the response.
-BaseController.prototype.update = function(req, reply, formatResponse) {
-  var reqData = this.formatRequestData(req);
-  formatResponse = typeof formatResponse === 'function' ? formatResponse : defaultFormatResponse;
-
-  var result = Promise.resolve().then(function() {
-    var record = req.pre.records.models[0];
-
-    record.set(reqData);
-    if (!record.hasChanged()) {
-      return record;
+    if (typeof formatResponse !== `function`) {
+      formatResponse = defaultFormatResponse;
     }
 
-    return record
-      .save(record.changed, { patch: true, method: 'update' });
-  })
-  .then(function (updatedState) {
-    return req.generateResponse(formatResponse(updatedState).toJSON()).code(200);
-  })
-  .catch(errors.generateErrorResponse);
-
-  reply(result);
-};
-
-// `formatResponse` is an optional processing function that can be passed
-// in to modify what is sent in the response body. If no function is
-// provided, the full model for the current method is used in the response.
-BaseController.prototype.create = function(req, reply, formatResponse) {
-  formatResponse = typeof formatResponse === 'function' ? formatResponse : defaultFormatResponse;
-
-  var result = this.Model
-    .forge(this.formatRequestData(req))
+    const result = this.Model
+    .forge(requestData)
     .save()
     .then(function(record) {
       if (!record) {
         throw Boom.notFound(null, {
-          error: 'Bookshelf error creating a resource'
+          error: `Bookshelf error creating a resource`
         });
       }
-      return req.generateResponse(formatResponse(record).toJSON())
-        .code(201);
+
+      return request.generateResponse(
+        formatResponse(record).toJSON()
+      ).code(201);
     })
-    .catch(errors.generateErrorResponse);
-  reply(result);
-};
+    .catch(Errors.generateErrorResponse);
 
-BaseController.prototype.delete = function(req, reply) {
-  var record = req.pre.records.models[0];
-
-  var result = Promise.resolve().then(function() {
-    return record.destroy();
-  })
-  .then(function() {
-    return req.generateResponse().code(204);
-  })
-  .catch(errors.generateErrorResponse);
-
-  reply(result);
-};
-
-BaseController.prototype.getAllAsMeta = function(req, reply) {
-  reply(req.generateResponse(req.pre.records.toJSON()));
-};
-
-// NOTE: creating the tarball for a project can be a lengthy process, so
-// we do it in multiple turns of the event loop, so that other requests
-// don't get blocked.
-BaseController.prototype.getAllAsTar = function(req, reply) {
-  var files = req.pre.records.models;
-  var tarStream = Tar.pack();
-  var model = this.Model;
-
-  function processFile(file) {
-    return model.query({
-      where: {
-        id: file.get('id')
-      },
-      columns: ['buffer']
-    }).fetch().then(function(model) {
-      return new Promise(function(resolve) {
-        setImmediate(function() {
-          tarStream.entry({ name: file.get('path') }, model.get('buffer'));
-          resolve();
-        });
-      });
-    });
+    reply(result);
   }
 
-  setImmediate(function() {
-    Promise.map(files, processFile, { concurrency: 2 })
-      .then(function() { tarStream.finalize(); })
-      .catch(errors.generateErrorResponse);
-  });
+  getOne(request, reply) {
+    const record = request.pre.records.models[0].toJSON();
 
-  // Normally this type would be application/x-tar, but IE refuses to
-  // decompress a gzipped stream when this is the type.
-  return reply(tarStream)
-    .header('Content-Type', 'application/octet-stream');
-};
+    reply(
+      request.generateResponse(
+        this.formatResponseData(record)
+      ).code(200)
+    );
+  }
 
+  getAll(request, reply) {
+    const records = request.pre.records.toJSON();
+
+    reply(
+      request.generateResponse(
+        records.map(this.formatResponseData)
+      )
+    );
+  }
+
+  getAllAsMeta(request, reply) {
+    reply(
+      request.generateResponse(
+        request.pre.records.toJSON()
+      )
+    );
+  }
+
+  // NOTE: creating the tarball for a project can be a lengthy process, so
+  // we do it in multiple turns of the event loop, so that other requests
+  // don't get blocked.
+  getAllAsTar(request, reply) {
+    const files = request.pre.records.models;
+    const tarStream = Tar.pack();
+    const Model = this.Model;
+
+    function processFile(file) {
+      return Model.query({
+        where: {
+          id: file.get(`id`)
+        },
+        columns: [`buffer`]
+      })
+      .fetch()
+      .then(function(model) {
+        return new Promise(function(resolve) {
+          setImmediate(function() {
+            tarStream.entry(
+              {
+                name: file.get(`path`)
+              },
+              model.get(`buffer`)
+            );
+
+            resolve();
+          });
+        });
+      });
+    }
+
+    setImmediate(function() {
+      Promise.map(files, processFile, { concurrency: 2 })
+      .then(function() { return tarStream.finalize(); })
+      .catch(Errors.generateErrorResponse);
+    });
+
+    // Normally this type would be application/x-tar, but IE refuses to
+    // decompress a gzipped stream when this is the type.
+    reply(tarStream)
+    .header(`Content-Type`, `application/octet-stream`);
+  }
+
+  // `formatResponse` is an optional processing function that can be passed
+  // in to modify what is sent in the response body. If no function is
+  // provided, the full model for the current method is used in the response.
+  update(request, reply, formatResponse) {
+    const requestData = this.formatRequestData(request);
+
+    if (typeof formatResponse !== `function`) {
+      formatResponse = defaultFormatResponse;
+    }
+
+    const result = Promise.resolve().then(function() {
+      const record = request.pre.records.models[0];
+
+      record.set(requestData);
+      if (!record.hasChanged()) {
+        return record;
+      }
+
+      return record.save(
+        record.changed,
+        {
+          patch: true,
+          method: `update`
+        }
+      );
+    })
+    .then(function(updatedState) {
+      return request.generateResponse(
+        formatResponse(updatedState).toJSON()
+      ).code(200);
+    })
+    .catch(Errors.generateErrorResponse);
+
+    reply(result);
+  }
+
+  delete(request, reply) {
+    const record = request.pre.records.models[0];
+
+    const result = Promise.resolve()
+    .then(function() { return record.destroy(); })
+    .then(function() { return request.generateResponse().code(204); })
+    .catch(Errors.generateErrorResponse);
+
+    reply(result);
+  }
+}
 
 module.exports = BaseController;
--- a/api/classes/base_model.js
+++ b/api/classes/base_model.js
@@ -1,13 +1,19 @@
-var Bookshelf = require('./database').Bookshelf;
+"use strict";
 
-var instanceProps = {
-  column: function(columnName, alias, escapeString) {
-    var prefix = escapeString ? '"' : '';
-    var column = prefix + this.tableName + prefix + '.' + prefix + columnName + prefix;
-    return alias ? column + ' AS ' + prefix + alias + prefix : column;
+const Bookshelf = require(`./database`).Bookshelf;
+
+const instanceProps = {
+  column(columnName, alias, escapeString) {
+    const delimiter = escapeString ? `"` : ``;
+    const tableName = `${delimiter}${this.tableName}${delimiter}`;
+    const fieldName = `${delimiter}${columnName}${delimiter}`;
+    const column = `${tableName}.${fieldName}`;
+
+    return alias ? `${column} AS ${delimiter}${alias}${delimiter}` : column;
   }
 };
-var classProps = {
+
+const classProps = {
   transaction: Bookshelf.transaction.bind(Bookshelf)
 };
 
--- a/api/classes/errors.js
+++ b/api/classes/errors.js
@@ -1,43 +1,48 @@
-var Boom = require('boom');
+"use strict";
 
-module.exports.attrs =  function(request, reply, source, error) {
-  var failedAttribute = '`' + error.data.details[0].path + '`';
+const Boom = require(`boom`);
 
-  // Look to see if the attribute was passed at all
-  if (error.data.details[0].message.indexOf('is required') !== -1) {
-    return reply(Boom.badRequest(failedAttribute + ' must be passed.', {
-      debug: true,
-      error: failedAttribute + ' must be passed.'
-    }));
-  }
+function generateBadRequest(param, isMissingError) {
+  const messageSuffix = isMissingError ? `must be passed.` : `invalid`;
+  const error = `${param} ${messageSuffix}`;
 
-  // Otherwise the type was invalid
-  return reply(Boom.badRequest(failedAttribute + ' invalid', {
+  return Boom.badRequest(error, {
     debug: true,
-    error: failedAttribute + ' invalid'
-  }));
-};
-
-module.exports.id = function(request, reply, source, error) {
-  return reply(Boom.badRequest('`id` invalid', {
-    debug: true,
-    error: '`id` invalid'
-  }));
-};
-
-module.exports.name = function(request, reply, source, error) {
-  return reply(Boom.badRequest('`name` invalid', {
-    debug: true,
-    error: '`name` invalid'
-  }));
-};
-
-module.exports.generateErrorResponse = function(e) {
-  if (e.isBoom) {
-    return e;
-  }
-
-  return Boom.badImplementation(null, {
-    error: e
+    error
   });
-};
+}
+
+class Errors {
+  static attrs(request, reply, source, error) {
+    const errorDetails = error.data.details[0];
+    const failedAttribute = `\`${errorDetails.path}\``;
+
+    // Look to see if the attribute was passed at all
+    if (errorDetails.message.indexOf(`is required`) !== -1) {
+      return reply(generateBadRequest(failedAttribute, true));
+    }
+
+    // Otherwise the type was invalid
+    return reply(generateBadRequest(failedAttribute));
+  }
+
+  /* eslint no-unused-vars: ["error", { "argsIgnorePattern": "^error$" }]*/
+  static id(request, reply, source, error) {
+    return reply(generateBadRequest(`\`id\``));
+  }
+
+  /* eslint no-unused-vars: ["error", { "argsIgnorePattern": "^error$" }]*/
+  static name(request, reply, source, error) {
+    return reply(generateBadRequest(`\`name\``));
+  }
+
+  static generateErrorResponse(error) {
+    if (error.isBoom) {
+      return error;
+    }
+
+    return Boom.badImplementation(null, { error });
+  }
+}
+
+module.exports = Errors;
--- a/api/classes/prerequisites.js
+++ b/api/classes/prerequisites.js
@@ -1,233 +1,256 @@
-var Boom = require('boom');
-var Promise = require('bluebird'); // jshint ignore:line
+"use strict";
 
-var Users = require('../modules/users/model');
-var errors = require('./errors');
+const Boom = require(`boom`);
+const Promise = require(`bluebird`);
 
-var prerequisites = {};
+const Users = require(`../modules/users/model`);
+const Errors = require(`./errors`);
 
-/**
- * confirmRecordExists(model[, mode, requestKey, databasekey])
- *
- * Returns a HAPI pre-req package configured to
- * to fetch all matching records of the passed `model`,
- * using data from the route parameters or the request payload
- * to build the query.
- *
- * @param {Object} model - Bookshelf model for querying
- * @param {String} mode - 'param' or 'payload' (if omitted, returns all records)
- * @param {String} requestKey - key of the param/payload
- * @param {String} databaseKey - key of the database model
- *
- * @return {Promise}
- */
-prerequisites.confirmRecordExists = function(model, config) {
-  config = config || {};
-  config.databaseKey = config.databaseKey || config.requestKey;
+class Prerequisites {
+  /**
+  * confirmRecordExists(model[, mode, requestKey, databasekey])
+  *
+  * Returns a HAPI pre-req package configured to
+  * to fetch all matching records of the passed `model`,
+  * using data from the route parameters or the request payload
+  * to build the query.
+  *
+  * @param {Object} model - Bookshelf model for querying
+  * @param {Object} config - Contains configuration options for the query:
+  *     mode {String} - 'param' or 'payload' (if omitted, returns all records)
+  *     requestKey {String} - key of the param/payload
+  *     databaseKey {String} - key of the database model
+  *
+  * @return {Promise} - Promise fullfilled when the record is found
+  */
+  static confirmRecordExists(model, config) {
+    config = config || {};
+    const dbKey = config.databaseKey || config.requestKey;
+    const requestKey = config.requestKey;
 
-  return {
-    assign: 'records',
-    method: function(req, reply) {
-      var queryOptions = {};
-      if (config.requestKey) {
-        queryOptions.where = {};
+    return {
+      assign: `records`,
+      method(req, reply) {
+        const queryOptions = {};
 
-        if (config.mode === 'param') {
-          queryOptions.where[config.databaseKey] = req.params[config.requestKey];
-        } else {
-          queryOptions.where[config.databaseKey] = req.payload[config.requestKey];
+        if (requestKey) {
+          queryOptions.where = {};
+
+          if (config.mode === `param`) {
+            queryOptions.where[dbKey] = req.params[requestKey];
+          } else {
+            queryOptions.where[dbKey] = req.payload[requestKey];
+          }
         }
-      }
 
-      var fetchOptions;
-      if (config.columns) {
-        fetchOptions = { columns: config.columns };
-      }
+        let fetchOptions = {};
 
-      var result = model.query(queryOptions)
+        if (config.columns) {
+          fetchOptions = { columns: config.columns };
+        }
+
+        const result = model
+        .query(queryOptions)
         .fetchAll(fetchOptions)
         .then(function(records) {
           if (records.length === 0) {
             throw Boom.notFound(null, {
               debug: true,
-              error: 'resource not found'
+              error: `resource not found`
             });
           }
 
           return records;
         })
-        .catch(errors.generateErrorResponse);
+        .catch(Errors.generateErrorResponse);
 
-      return reply(result);
-    }
-  };
-};
+        return reply(result);
+      }
+    };
+  }
 
-/**
- * validateUser()
- *
- * Ensures that the user sending the request exists in the
- * current context. This means that the user should have hit the
- * /users/login route first
- *
- * @return {Promise}
- */
-prerequisites.validateUser = function() {
-  return {
-    assign: 'user',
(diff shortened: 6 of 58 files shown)


api/classes/base_model.js
```diff
@@ -1,13 +1,19 @@
...
-    var prefix = escapeString ? '"' : '';
-    var column = prefix + this.tableName + prefix + '.' + prefix + columnName + prefix;
-    return alias ? column + ' AS ' + prefix + alias + prefix : column;
+const Bookshelf = require(`./database`).Bookshelf;
+
+const instanceProps = {
+  column(columnName, alias, escapeString) {
+    const prefix = escapeString ? `"` : ``;
+    const column = `${prefix}${this.tableName}${prefix}` +
+                   `.` +
```
cadecairos: Does this `.` really need its own line still?

gideonthomas: Ah good catch!

Pomax: a better question is why are you using templates *and* string concatenation?

adaptors/extensions.js
```diff
@@ -1,59 +1,63 @@
...
-  var error = data && data.error;
+  const data = response.data;
+  const error = data && data.error;
 
   // Prefer the error object stack over the
   // boom object stack
-  var stack = error && error.stack || req.response.stack;
+  const stack = error && error.stack || response.stack;
+
+  let logLevel = `error`;
```
Pomax: this is already function scope, why `let` instead of `var`? (the additional block scoping for `let` will do nothing here)

adaptors/plugins.js
```diff
@@ -1,4 +1,6 @@
+"use strict";
+
 module.exports = [
-  require('lout'),
-  require('hapi-auth-bearer-token')
+  require(`lout`),
```
Pomax: is there a reason we're exporting an array here, rather than an object? This feels brittle (why would "lout" be array[0], for instance? Whereas returnval.lout being lout makes a whole bunch of sense)

api/classes/base_controller.js
```diff
@@ -11,133 +13,165 @@ function defaultFormatResponse(model) {
...
+class BaseController {
+  constructor(Model) {
+    this.Model = Model;
+  }
 
-BaseController.prototype.formatRequestData = function(req) {
-  // Abstract base method,
-  // formats data for database entry
-};
+  /* eslint no-unused-vars: ["error", { "argsIgnorePattern": "^request$" }]*/
```
Pomax: what is this preventing?

api/classes/base_controller.js
```diff
@@ -11,133 +13,165 @@ function defaultFormatResponse(model) {
...
+  formatRequestData(request) {
+    // Abstract base method; formats data for database entry
```
Pomax: instead of an empty function, it might be worth having a `throw new Error("function not implemented by subclass");`

api/classes/base_controller.js
```diff
@@ -11,133 +13,165 @@ function defaultFormatResponse(model) {
...
-    var record = req.pre.records.models[0];
-
-    record.set(reqData);
-    if (!record.hasChanged()) {
-      return record;
-    }
+  formatResponseData(model) {
+    // Base method to format data that will be sent in the response
+    // By default, it returns the model as is
+    return model;
```
Pomax: does this make sense, is a passthrough ever actually used?

api/classes/base_controller.js
```diff
@@ -11,133 +13,165 @@ function defaultFormatResponse(model) {
...
+  }
+
+  // `formatResponse` is an optional processing function that can be passed
+  // in to modify what is sent in the response body. If no function is
+  // provided, the full model for the current method is used in the response.
+  create(request, reply, formatResponse) {
+    const requestData = this.formatRequestData(request);
+
+    formatResponse = typeof formatResponse === `function` ? formatResponse : defaultFormatResponse;
```
Pomax: this seems an odd thing to do: what are we testing here? why not be more explicit here and say:
```
if (typeof formatResponse !== "function) {
  formatResponse = defaultFormatResponse
}
```
rather than a blind reassigment that might reassign to the same reference.

api/classes/base_controller.js
```diff
@@ -11,133 +13,165 @@ function defaultFormatResponse(model) {
...
+    .header(`Content-Type`, `application/octet-stream`);
+  }
+
+  // `formatResponse` is an optional processing function that can be passed
+  // in to modify what is sent in the response body. If no function is
+  // provided, the full model for the current method is used in the response.
+  update(request, reply, formatResponse) {
+    const requestData = this.formatRequestData(request);
+
+    formatResponse = typeof formatResponse === `function` ? formatResponse : defaultFormatResponse;
```
Pomax: same note - this might be better as an explicit "don't do anything unless typeof is not function"

api/classes/base_controller.js
```diff
@@ -11,133 +13,165 @@ function defaultFormatResponse(model) {
...
+      );
+    })
+    .then(function(updatedState) {
+      return request.generateResponse(
+        formatResponse(updatedState).toJSON()
+      ).code(200);
+    })
+    .catch(Errors.generateErrorResponse);
+
+    reply(result);
```
Pomax: this is a little curious: why are we returning result, usually done so that promise handling can be done elsewhere, while also having promise handling tacked on in this function?

api/classes/base_model.js
```diff
@@ -1,13 +1,19 @@
...
-var instanceProps = {
-  column: function(columnName, alias, escapeString) {
-    var prefix = escapeString ? '"' : '';
-    var column = prefix + this.tableName + prefix + '.' + prefix + columnName + prefix;
-    return alias ? column + ' AS ' + prefix + alias + prefix : column;
+const Bookshelf = require(`./database`).Bookshelf;
+
+const instanceProps = {
+  column(columnName, alias, escapeString) {
+    const prefix = escapeString ? `"` : ``;
```
Pomax: this is not a `prefix`, this is a `delimiter`

api/classes/errors.js
```diff
@@ -1,43 +1,70 @@
...
-  // Look to see if the attribute was passed at all
-  if (error.data.details[0].message.indexOf('is required') !== -1) {
-    return reply(Boom.badRequest(failedAttribute + ' must be passed.', {
-      debug: true,
-      error: failedAttribute + ' must be passed.'
-    }));
+class Errors {
+  static attrs(request, reply, source, error) {
+    const errorDetails = error.data.details[0];
+    const failedAttribute = `\`${errorDetails.path}\``;
```
Pomax: do these need to be backticks inside the string?

api/classes/errors.js
```diff
@@ -1,43 +1,70 @@
...
+        }
+      )
+    );
+  }
+
+  /* eslint no-unused-vars: ["error", { "argsIgnorePattern": "^error$" }]*/
+  static id(request, reply, source, error) {
+    return reply(
+      Boom.badRequest(
+        `\`id\` invalid`,
```
Pomax: this is a duplicated string (L39 and L42), let's `const` that out

api/classes/errors.js
```diff
@@ -1,43 +1,70 @@
...
-};
-
-module.exports.generateErrorResponse = function(e) {
-  if (e.isBoom) {
-    return e;
+  /* eslint no-unused-vars: ["error", { "argsIgnorePattern": "^error$" }]*/
+  static name(request, reply, source, error) {
+    return reply(
+      Boom.badRequest(
+        `\`name\` invalid`,
```
Pomax: DRY - repeated in the next arg

api/classes/prerequisites.js
```diff
@@ -1,233 +1,255 @@
...
+  * @param {Object} model - Bookshelf model for querying
+  * @param {Object} config - Contains configuration options for the query:
+  *     mode {String} - 'param' or 'payload' (if omitted, returns all records)
+  *     requestKey {String} - key of the param/payload
+  *     databaseKey {String} - key of the database model
+  *
+  * @return {Promise} - Promise fullfilled when the record is found
+  */
+  static confirmRecordExists(model, config) {
+    config = config || {};
```
Pomax: we're in ES6, which means we can do this in the parameter signature instead

api/classes/prerequisites.js
```diff
@@ -1,233 +1,255 @@
...
+    config.databaseKey = config.databaseKey || config.requestKey;
```
Pomax: does this need to be a config rewrite? (i.e. persist beyond this function). If not, let's do this with a `const` instead,
```
const configKey = config.databaseKey || config.requestKey;
const requestKey = config.requestKey;
...
      queryOptions.where[configKey] = req.params[requestKey];
...

```

api/classes/prerequisites.js
```diff
@@ -1,233 +1,255 @@
...
 
-      var fetchOptions;
-      if (config.columns) {
-        fetchOptions = { columns: config.columns };
-      }
+        let fetchOptions;
 
-      var result = model.query(queryOptions)
+        if (config.columns) {
+          fetchOptions = { columns: config.columns };
```
Pomax: this reads as odd process: what happens when using `fetchOptions = {}; if (config.columns) { fetchOptions.columns = config.columns; }`?

api/classes/publisher.js
```diff
@@ -1,411 +1,412 @@
...
-var UserQueries = require('../modules/users/model').prototype.queries();
-var ProjectQueries = Projects.prototype.queries();
-var PublishedProjectQueries = PublishedProjects.prototype.queries();
-var PublishedFileQueries = require('../modules/publishedFiles/model').prototype.queries();
+const Projects = require(`../modules/projects/model`);
+const PublishedProjects = require(`../modules/publishedProjects/model`);
 
-var reject  = Promise.reject;
+// SQL Query Generators
+const usersQueryBuilder = require(`../modules/users/model`).prototype.queryBuilder();
```
Pomax: this looks realy bad - if we're in ES6 why are we pulling things from prototypes instead of using static functions?

api/classes/publisher.js
```diff
@@ -1,411 +1,412 @@
...
 // of the path to return a fully uri safe path
 function uriSafe(path) {
-  if (path === '/') {
-    return '/';
+  if (path === `/`) {
+    return `/`;
   }
 
-  var uriSafePath = '';
+  let uriSafePath = ``;
```
Pomax: function scoped: this really is a `var`, not a `let`

api/classes/publisher.js
```diff
@@ -1,411 +1,412 @@
...
 
-  while (path !== '/') {
-    uriSafePath = Path.join('/', encodeURIComponent(Path.basename(path)), uriSafePath);
+  while(path !== `/`) {
+    uriSafePath = Path.join(
+      `/`,
```
Pomax: we're using the `\`/\`` string a lot here - let's catch it as a `const rootURL` or something?

api/classes/publisher.js
```diff
@@ -1,411 +1,412 @@
...
+        return;
       }
 
-      return remove(fileRoot + oldPath)
-      .then(function() {
-        return updateModelAndUpload(publishedFile);
+      return Promise.map(publishedFiles, function(publishedFile) {
+        const oldPath = publishedFile.oldPath;
+
+        delete publishedFile.oldPath;
```
Pomax: this is a super expensive operation (due to JIT invalidation), do we *need* to delete this, or can we just set it to `undefined`?

api/classes/publisher.js
```diff
@@ -1,411 +1,412 @@
...
+    .then(() => this.fetchUserForProject())
+    .then(() => this.createOrUpdatePublishedProject())
+    .then(() => this.setRemixDataForPublishedProject())
+    .then(() => this.uploadNewFiles())
+    .then(() => this.uploadModifiedFiles())
+    .then(() => this.deleteOldFiles())
+    .then(() => this.updateProjectDetails())
+    .then(() => {
+      log.info(
+        `Publish for ${this.user.name} - ` +
```
Pomax: template strings with string concatenation feels really wrong. Doesn't `log` already insert a space between multiple arguments?

api/classes/publisher.js
```diff
@@ -1,411 +1,412 @@
...
-    project: project.toJSON()
-  };
+  unpublish() {
+    return Promise.resolve()
+    .then(() => this.fetchUserForProject())
+    .then(() => this.fetchPublishedProject())
+    .then(() => this.deletePublishedFiles())
+    .then(() => this.deletePublishedProject())
+    .then(() => {
+      log.info(
```
Pomax: template strings and string concatenation

api/modules/files/model.js
```diff
@@ -1,28 +1,31 @@
...
+  tableName: `files`,
+  project() {
+    return this.belongsTo(ProjectsModel);
   },
-  user: function() {
-    return this.belongsTo(Users).through(Projects);
+  user() {
+    // We require in the function as opposed to adding a top-level require
+    // to resolve the circular dependencies between models
+    return this.belongsTo(require(`../users/model`)).through(ProjectsModel);
```
Pomax: requires are cached - why are we doing a late require here?

api/modules/files/routes/create.js
```diff
@@ -1,29 +1,32 @@
...
     },
     pre: [
-      prereqs.trackTemporaryFile(),
-      prereqs.validateCreationPermission('project_id', Project)
+      Prerequistes.trackTemporaryFile(),
+      Prerequistes.validateCreationPermission(`project_id`, ProjectsModel)
     ],
-    handler: controller.create.bind(controller),
-    description: 'Create a new file object.',
+    handler: filesController.create.bind(filesController),
```
Pomax: `args => filesController(...args)`

api/modules/files/routes/delete.js
```diff
@@ -1,30 +1,32 @@
...
+        requestKey: `id`
       }),
-      prereq.validateUser(),
-      prereq.validateOwnership()
+      Prerequistes.validateUser(),
+      Prerequistes.validateOwnership()
     ],
-    handler: controller.delete.bind(controller),
-    description: 'Delete a single file object based on `id`.',
+    handler: filesController.delete.bind(filesController),
```
Pomax: `handler: args => filesController.delete(...args),` although it feels odd that this can't just be a static function and needs explicit tying to a specific fileController instance. It suggests we might have unnecessary data capturing in the fileController class.

api/modules/files/routes/read.js
```diff
@@ -1,25 +1,27 @@
...
+        requestKey: `id`
       }),
-      prereqs.validateUser(),
-      prereqs.validateOwnership()
+      Prerequisites.validateUser(),
+      Prerequisites.validateOwnership()
     ],
-    handler: controller.getOne.bind(controller),
-    description: 'Retrieve a single file object based on `id`.',
+    handler: filesController.getOne.bind(filesController),
```
Pomax: `handler: args => filesController.getOne(...args),` although it feels odd that this can't just be a static function and needs explicit tying to a specific fileController instance. It suggests we might have unnecessary data capturing in the fileController class.

api/modules/files/routes/read.js
```diff
@@ -28,19 +30,19 @@ module.exports = [{
...
+        requestKey: `project_id`
       }),
-      prereqs.validateUser(),
-      prereqs.validateOwnership()
+      Prerequisites.validateUser(),
+      Prerequisites.validateOwnership()
     ],
-    handler: controller.getAll.bind(controller),
-    description: 'Retrieve a collection of file objects that belong to a single project object, based on `project_id`.',
+    handler: filesController.getAll.bind(filesController),
```
Pomax: `handler: args => filesController.getAll(...args),` although it feels odd that this can't just be a static function and needs explicit tying to a specific fileController instance. It suggests we might have unnecessary data capturing in the fileController class.

api/modules/files/routes/read.js
```diff
@@ -49,20 +51,20 @@ module.exports = [{
...
+        columns: [`id`, `project_id`, `path`]
       }),
-      prereqs.validateUser(),
-      prereqs.validateOwnership()
+      Prerequisites.validateUser(),
+      Prerequisites.validateOwnership()
     ],
-    handler: controller.getAllAsMeta.bind(controller),
-    description: 'Retrieve a collection of file objects that belong to a single project object, based on `project_id`.',
+    handler: filesController.getAllAsMeta.bind(filesController),
```
Pomax: `handler: args => filesController.getAllAsMeta(...args),` although it feels odd that this can't just be a static function and needs explicit tying to a specific fileController instance. It suggests we might have unnecessary data capturing in the fileController class.

api/modules/files/routes/read.js
```diff
@@ -71,21 +73,21 @@ module.exports = [{
...
       }),
-      prereqs.validateUser(),
-      prereqs.validateOwnership()
+      Prerequisites.validateUser(),
+      Prerequisites.validateOwnership()
     ],
-    handler: controller.getAllAsTar.bind(controller),
-    description: 'Retrieve a tar file containing a collection of file objects that belong to a ' +
-    'single project object, based on `project_id`.',
+    handler: filesController.getAllAsTar.bind(filesController),
```
Pomax: `handler: args => filesController.getAllAsTar(...args),` although it feels odd that this can't just be a static function and needs explicit tying to a specific fileController instance. It suggests we might have unnecessary data capturing in the fileController class.

api/modules/files/routes/read.js
```diff
@@ -71,21 +73,21 @@ module.exports = [{
...
+    description: `Retrieve a tar file containing a collection of file objects that belong to a ` +
```
Pomax: template strings and string concatenation - we don't chop up the description above, so we'll want to pick a style here. Either chop all strings, or don't chop any.

api/modules/files/routes/update.js
```diff
@@ -1,33 +1,35 @@
...
+        requestKey: `id`
       }),
-      prereqs.validateUser(),
-      prereqs.validateOwnership()
+      Prerequisites.validateUser(),
+      Prerequisites.validateOwnership()
     ],
-    handler: controller.update.bind(controller),
-    description: 'Update a single file object based on `id`.',
+    handler: filesController.update.bind(filesController),
```
Pomax: `handler: args => filesController.update(...args),` although it feels odd that this can't just be a static function and needs explicit tying to a specific fileController instance. It suggests we might have unnecessary data capturing in the fileController class.

api/modules/projects/controller.js
```diff
@@ -1,144 +1,158 @@
...
-};
+class PublishedFilesCleanup {
+  constructor(project) {
+    this.project = project;
+  }
 
-controller.unpublishProject = function(req, reply) {
-  var result = Promise.resolve().then(function() {
-    var record = req.pre.records.models[0];
+  _fetchPublishedFiles(fileModel) {
```
Pomax: why underscored names?

api/modules/projects/controller.js
```diff
@@ -1,144 +1,158 @@
...
+}
+
+class ProjectsController extends BaseController {
+  constructor() {
+    super(ProjectsModel);
+  }
+
+  formatRequestData(request) {
+    const now = new Date();
+    const data = {
```
Pomax: this feels like `const data = request.payload`

api/modules/projects/controller.js
```diff
@@ -1,144 +1,158 @@
...
+      date_created: now,
+      date_updated: now,
+      readonly: request.payload.readonly,
+      client: request.payload.client
+    };
+
+    // If it is an update request
+    if (request.params.id) {
+      data.id = parseInt(request.params.id, 10);
+      delete data.date_created;
```
Pomax: very expensive op, does setting it to `undefined` work?

api/modules/projects/routes/create.js
```diff
@@ -1,21 +1,23 @@
...
+  method: `POST`,
+  path: `/projects`,
   config: {
     pre: [
-      prereqs.validateCreationPermission()
+      Prerequisites.validateCreationPermission()
     ],
-    handler: controller.create.bind(controller),
-    description: 'Create a new project object.',
+    handler: projectsController.create.bind(projectsController),
```
Pomax: `bind` in ES6 is a sign of trouble

api/modules/projects/routes/delete.js
```diff
@@ -1,30 +1,32 @@
...
+        requestKey: `id`
       }),
-      prereqs.validateUser(),
-      prereqs.validateOwnership()
+      Prerequisites.validateUser(),
+      Prerequisites.validateOwnership()
     ],
-    handler: controller.delete.bind(controller),
-    description: 'Delete a single project object based on `id`.',
+    handler: projectsController.delete.bind(projectsController),
```
Pomax: bind

api/modules/projects/routes/publish.js
```diff
@@ -1,36 +1,38 @@
...
+        requestKey: `id`
       }),
-      prereqs.validateUser(),
-      prereqs.validateOwnership()
+      Prerequisites.validateUser(),
+      Prerequisites.validateOwnership()
     ],
-    handler: controller.publishProject.bind(controller),
-    description: 'Publish a project.'
+    handler: projectsController.publish.bind(projectsController),
```
Pomax: bind

api/modules/projects/routes/publish.js
```diff
@@ -1,36 +1,38 @@
...
+        requestKey: `id`
       }),
-      prereqs.validateUser(),
-      prereqs.validateOwnership()
+      Prerequisites.validateUser(),
+      Prerequisites.validateOwnership()
     ],
-    handler: controller.unpublishProject.bind(controller),
-    description: 'Unpublish a project.'
+    handler: projectsController.unpublish.bind(projectsController),
```
Pomax: bind

api/modules/projects/routes/read.js
```diff
@@ -1,51 +1,53 @@
...
+        requestKey: `id`
       }),
-      prereqs.validateUser(),
-      prereqs.validateOwnership()
+      Prerequisites.validateUser(),
+      Prerequisites.validateOwnership()
     ],
-    handler: controller.getOne.bind(controller),
-    description: 'Retrieve a single project object based on `id`.',
+    handler: projectsController.getOne.bind(projectsController),
```
Pomax: bind

api/modules/projects/routes/update.js
```diff
@@ -1,32 +1,34 @@
...
+        requestKey: `id`
       }),
-      prereqs.validateUser(),
-      prereqs.validateOwnership()
+      Prerequisites.validateUser(),
+      Prerequisites.validateOwnership()
     ],
-    handler: controller.update.bind(controller),
-    description: 'Update a single project object based on `id`.',
+    handler: projectsController.update.bind(projectsController),
```
Pomax: bind

api/modules/publishedFiles/model.js
```diff
@@ -1,113 +1,144 @@
...
+    .query()
+    .where(this.context.column(`published_id`), publishedId)
+    .del();
+  }
+}
+
+const instanceProps = {
+  tableName: `publishedFiles`,
+  project() {
+    return this.belongsTo(require(`../publishedProjects/model`));
```
Pomax: can we move this late require to not be a late require?

api/modules/publishedFiles/routes/read.js
```diff
@@ -1,24 +1,26 @@
...
-        mode: 'param',
-        requestKey: 'id'
+      Prerequisites.confirmRecordExists(PublishedFilesModel, {
+        mode: `param`,
+        requestKey: `id`
       })
     ],
-    handler: controller.getOne.bind(controller),
-    description: 'Retrieve a single published file object based on `id`.',
+    handler: publishedFilesController.getOne.bind(publishedFilesController),
```
Pomax: bind

api/modules/publishedFiles/routes/read.js
```diff
@@ -27,19 +29,19 @@ module.exports = [{
...
-        requestKey: 'published_id'
+      Prerequisites.confirmRecordExists(PublishedFilesModel, {
+        mode: `param`,
+        requestKey: `published_id`
       })
     ],
-    handler: controller.getAll.bind(controller),
-    description: 'Retrieve a collection of published file objects that belong to a single published project object, ' +
-    'based on `published_id`.',
+    handler: publishedFilesController.getAll.bind(publishedFilesController),
```
Pomax: bind

api/modules/publishedFiles/routes/read.js
```diff
@@ -48,20 +50,20 @@ module.exports = [{
...
+      Prerequisites.confirmRecordExists(PublishedFilesModel, {
+        mode: `param`,
+        requestKey: `published_id`,
+        columns: [`id`, `published_id`, `file_id`, `path`]
       })
     ],
-    handler: controller.getAllAsMeta.bind(controller),
-    description: 'Retrieve a collection of publishedFile objects that belong to a single project object, based on ' +
-    '`published_id`. Omits `buffer` attribute.',
+    handler: publishedFilesController.getAllAsMeta.bind(publishedFilesController),
```
Pomax: bind

api/modules/publishedFiles/routes/read.js
```diff
@@ -70,20 +72,20 @@ module.exports = [{
...
+      Prerequisites.confirmRecordExists(PublishedFilesModel, {
+        mode: `param`,
+        requestKey: `published_id`,
+        columns: [`id`, `path`]
       })
     ],
-    handler: controller.getAllAsTar.bind(controller),
-    description: 'Retrieve a collection of publishedFile objects that belong to a single project object, based on ' +
-    '`published_id`. Omits `buffer` attribute.',
+    handler: publishedFilesController.getAllAsTar.bind(publishedFilesController),
```
Pomax: bind

api/modules/publishedProjects/routes/read.js
```diff
@@ -27,14 +29,14 @@ module.exports = [{
...
+  path: `/publishedProjects`,
   config: {
     auth: false,
     pre: [
-      prereqs.confirmRecordExists(Model)
+      Prerequisites.confirmRecordExists(PublishedProjectsModel)
     ],
-    handler: controller.getAll.bind(controller),
-    description: 'Retrieve a collection of published project objects.'
+    handler: publishedProjectsController.getAll.bind(publishedProjectsController),
```
Pomax: bind

api/classes/base_model.js
gideonthomas: I use both because of character length per line constraints.
Do you know how to put them on separate lines with backticks so that it won't consider the newlines @Pomax?

adaptors/extensions.js
gideonthomas: yeah you're right out here it makes no difference, but I'm not sure what convention we're following?
Are we choosing:
* Use `let` and only use `var` when we have to (variable hoisting) OR
* Use `var` and only use `let` when we have to (block scoping)

adaptors/plugins.js
gideonthomas: I think this is how it is passed into Hapi (directly from what is exported)? I'm not sure, I didn't change the logic here.

api/classes/base_controller.js
gideonthomas: eslint complains about the `request` argument in the function signature because it is not used in the function. I keep it in there mainly so that if you decide to override the function (since this function is really an abstract method), you know what param to expect in the function.

api/classes/base_controller.js
gideonthomas: Ah good idea!

api/classes/base_controller.js
gideonthomas: yep it is always called and this default implementation is actually called about 60% of the time.

api/classes/errors.js
gideonthomas: Well I think this was meant for delimiting the path from the rest of the text in a log. It seems to be a convention followed throughout the code.

api/classes/prerequisites.js
gideonthomas: not supported by node 4

api/classes/publisher.js
gideonthomas: Because the exports for this require exposes something that is generated by an external library (viz. Bookshelf) which does not use ES6.

api/classes/publisher.js
gideonthomas: Interesting! Yeah I think I'll set it to `null` since it will practically have the same effect.

api/modules/files/model.js
gideonthomas: yes...it's a very annoying bug! Requiring this at runtime fixes a circular dependency so that the exports happens first before the require.

api/modules/files/routes/create.js
gideonthomas: the spread operator is not supported (without the `--harmony` flag) in node 4.4

api/modules/publishedProjects/routes/remix.js
```diff
@@ -1,20 +1,22 @@
...
+      Prerequisites.confirmRecordExists(PublishedProjectsModel, {
+        mode: `param`,
+        requestKey: `id`
       }),
-      prereqs.validateUser()
+      Prerequisites.validateUser()
     ],
-    handler: controller.remix.bind(controller),
-    description: 'Create a new project object.'
+    handler: publishedProjectsController.remix.bind(publishedProjectsController),
```
Pomax: bind

api/modules/users/routes/delete.js
```diff
@@ -1,30 +1,32 @@
...
+        requestKey: `id`
       }),
-      prereqs.validateUser(),
-      prereqs.validateOwnership()
+      Prerequisites.validateUser(),
+      Prerequisites.validateOwnership()
     ],
-    handler: controller.delete.bind(controller),
-    description: 'Delete a user object based on `id`.',
+    handler: usersController.delete.bind(usersController),
```
Pomax: bind

api/modules/users/routes/login.js
```diff
@@ -1,17 +1,19 @@
...
 
 module.exports = [{
-  method: 'POST',
-  path: '/users/login',
+  method: `POST`,
+  path: `/users/login`,
   config: {
-    handler: controller.login.bind(controller),
-    description: 'Retrieve the user with the passed username, creating if necessary.',
+    handler: usersController.login.bind(usersController),
```
Pomax: bind

api/modules/users/routes/read.js
```diff
@@ -1,30 +1,32 @@
...
+        requestKey: `id`
       }),
-      prereqs.validateUser(),
-      prereqs.validateOwnership()
+      Prerequisites.validateUser(),
+      Prerequisites.validateOwnership()
     ],
-    handler: controller.getOne.bind(controller),
-    description: 'Retrieve a single user object based on `id`.',
+    handler: usersController.getOne.bind(usersController),
```
Pomax: bind

api/modules/users/routes/update.js
```diff
@@ -1,32 +1,34 @@
...
+        requestKey: `id`
       }),
-      prereqs.validateUser(),
-      prereqs.validateOwnership()
+      Prerequisites.validateUser(),
+      Prerequisites.validateOwnership()
     ],
-    handler: controller.update.bind(controller),
-    description: 'Update a user object based on `id`.',
+    handler: usersController.update.bind(usersController),
```
Pomax: bind

lib/remix.js
```diff
@@ -1,33 +1,43 @@
...
+const REMIX_SCRIPT = process.env.REMIX_SCRIPT;
+
+Hoek.assert(REMIX_SCRIPT, `Must define location of the remix script`);
+
+const remixUrl = Url.parse(REMIX_SCRIPT);
+const slashes = remixUrl.slashes ? `//` : ``;
 
 function injectMetadata(html, metadata) {
-  var metaTags = '';
+  let metaTags = ``;
```
Pomax: function scoped var = `var`, not `let`

lib/tokenValidator.js
```diff
@@ -1,38 +1,45 @@
...
+  }, function(error, response, body) {
+    if (error) {
+      return callback(error);
     }
 
-    if ( resp.statusCode !== 200 ) {
-      if ( resp.statusCode === 401 ) {
+    if (response.statusCode !== 200) {
+      if (response.statusCode === 401) {
         return callback(null, false);
```
Pomax: not part of your rewrite but: this feels like it should have just been `callback()` =)

lib/utils.js
```diff
@@ -20,48 +21,58 @@ var dateTracker = {
...
+  }
 
-      if (isDate(updated)) {
-        if (isModel) {
-          data.set('date_updated', updated.toISOString());
-        } else {
-          data.date_updated = updated.toISOString();
-        }
-      }
+  static extractDate(data, property, isModel) {
```
Pomax: this is not what the function does. It's a generic `get`, so calling it `extractDate` may be misleading for future readers of this code.

lib/utils.js
```diff
@@ -20,48 +21,58 @@ var dateTracker = {
...
+    return isModel ? data.get(property) : data[property];
+  }
+
+  static convertToISOStrings(data) {
+    const created = DateTracker.extractDate(data, `date_created`);
```
Pomax: what inputs are you provisioning for here? It looks like the function starts uncertain what it's working on, but then later in the function sets values as if `data` is always a generic object

package.json
```diff
@@ -24,7 +23,7 @@
   },
   "homepage": "https://github.com/mozilla/publish.webmaker.org",
   "engines": {
-    "node": "4.2.x"
+    "node": "4.4.5"
```
Pomax: patch versions should be fine, `4.4.x` should be reasonable

scripts/export.js
```diff
@@ -1,115 +1,122 @@
...
+
+const Promise = require(`bluebird`);
+const program = require(`commander`);
+const PgClient = require(`pg`).Client;
+const Tar = require(`tar-stream`);
+const colors = require(`colors`);
+const fs = Promise.promisifyAll(require(`fs`));
+const path = require(`path`);
+
+let client;
```
Pomax: if you use `let`, assign it immediately. In this case, as top level variables without a block scope, these should just be `var`

scripts/export.js
```diff
@@ -1,115 +1,122 @@
...
+  help: `cyan`,
+  warn: `yellow`,
+  debug: `blue`,
+  error: `red`
 });
 
 function getMessage() {
-  var message = '';
+  var message = ``;
   Array.prototype.forEach.call(arguments, function(arg) {
```
Pomax: `Array.from(arguments).forEach(....)`

scripts/export.js
```diff
@@ -1,115 +1,122 @@
...
 }
 
-var log = {
-  error: function() {
-    var message = getMessage.apply(null, arguments);
-    console.error(('ERROR: ' + message).error);
-    console.error('\n\n');
+class Log {
+  static error() {
+    const message = getMessage.apply(null, arguments);
```
Pomax: apply in ES6 is highly questionable - what are you trying to do here? `getMessage(...arguments)` should do the trick already

scripts/export.js
```diff
@@ -1,115 +1,122 @@
...
+    console.error((`ERROR: ${message}`).error);
+    console.error(`\n\n`);
     process.exit(1);
-  },
-  info: function() {
-    var message = getMessage.apply(null, arguments);
+  }
+
+  static info() {
+    const message = getMessage.apply(null, arguments);
```
Pomax: see above

scripts/export.js
```diff
@@ -1,115 +1,122 @@
...
     console.log(message.verbose);
-  },
-  success: function() {
-    var message = getMessage.apply(null, arguments);
+  }
+
+  static success() {
+    const message = getMessage.apply(null, arguments);
```
Pomax: see above

scripts/export.js
```diff
@@ -1,115 +1,122 @@
...
 }
 
 function writeProjectDetailsFile(project) {
-  return fs.writeFileAsync(path.join(destination, 'project.txt'), JSON.stringify(project, null, 4))
-  .then(log.info.bind(log, 'Created project details file...'))
+  return fs.writeFileAsync(
+    path.join(destination, `project.txt`),
+    JSON.stringify(project, null, 4)
+  )
+  .then(Log.info.bind(Log, `Created project details file...`))
```
Pomax: bind

scripts/export.js
```diff
@@ -143,26 +150,26 @@ client.connectAsync()
...
+  const table = program.published ? `"publishedFiles"` : `"files"`;
+  const fk = program.published ? `"published_id"` : `"project_id"`;
+  const id = program.published ? project.published_id : projectID;
+  const cols = [ `path`, `buffer` ];
 
   return client.queryAsync(
-    'SELECT ' + '"' + cols.join('", "') + '" ' +
-    'FROM ' + table + ' ' +
-    'WHERE ' + fk + ' = $1',
+    `SELECT "` + cols.join('", "') + `" ` +
```
Pomax: you're using template strings.
```
`SELECT "${cols.join('", "')}"`
```

server.js
```diff
@@ -97,14 +97,15 @@ server.register({ register: require('./api') }, function(err) {
...
+  let port = endpoint.match(/:(\d+)/);
+
   if (!port) {
     return;
   }
 
-  var port = parseInt(port[1]);
-  require('mox-server').runServer(port, function() {
-    console.log('running mox server on port', port);
+  port = parseInt(port[1], 10);
```
Pomax: parseInt actually changed in ES6, it has been redefined to parse base 10 unless otherwise specified, unlike ES5, where this was not the case. As such, you don't need the `parseInt( ..., 10)` change here.

gideonthomas: interesting! This still fails with eslint though (mofo-style specifically I think)

Pomax: that might be either a bug in eslint or an incompletion in mofo-style, can we find out which?

api/classes/base_model.js
Pomax: I'd argue not to split, but that's because I use widescreen text editors, and if a line is code, in which which the placement matters, I don't mess with that placement. If a string really is *that* long, that's usually a sign of trying to do too much in the wrong place. For instance:
```
let tableName = `${prefix}${this.tableName}${prefix}`;
let fieldName = `${prefix}${columnName}${prefix}`;
const column = `${tableName}.${fieldName}`;
```
boom: clean, short, sorted.

adaptors/extensions.js
Pomax: there shouldn't be a "choice", their function is clearly defined in the spec. If you need a function level variable, you use `var`, if you need a block scoped variable, you use `let`, so that should be the guiding principle: does it need function scoping? use `var`. Does it need to only exist for as long as the current block is active? use `let`.

api/classes/base_controller.js
Pomax: if we make it involve `request`, too, we can remove the eslint override as well

api/classes/prerequisites.js
Pomax: hmm, then we need to change the description of this PR, because this is not a proper rewrite yet then.

api/modules/files/routes/create.js
Pomax: frkcn node 4 =)

api/modules/projects/controller.js
gideonthomas: supposed to signify that these methods aren't really meant to be called outside this class. They are internal implementation methods.

api/modules/projects/controller.js
gideonthomas: it also comes with the published_id which I don't think we want to persist.

api/modules/projects/controller.js
Pomax: Hm, that's not how JS works though, feels like applying a Python convention to a JS codebase. I'd recommend keeping them named normally and simply documenting them as "not meant for calls outside of this class", or literally making them external functions in the same closure context so that calls from within the class code work, but nothing outside it can see them.

api/modules/projects/controller.js
Pomax: noted

server.js
gideonthomas: I will override this rule in this repo's eslint config file since we're in a node env which does not need the radix.

api/modules/projects/controller.js
gideonthomas: unfortunately I can't do it out here since it looks like Bookshelf uses `hasOwnProperty` to determine if the key exists or not.

adaptors/plugins.js
cadecairos: yes, I believe the exported array is directly passed to hapi's register function

api/modules/projects/controller.js
cadecairos: +1 to Pomax's comment.

api/modules/projects/controller.js
cadecairos: we're referencing `payload` thrugh `request` a lot here. it'd be easier on the eyes if we stored a reference to `payload`

gideonthomas:
@cadecairos says merge away!

@Pomax and @cadecairos Thank you very much for the reviews! Really appreciate it!

PR #231, opened 2016-07-07
Fix regression in ES6 re-write that replaces delete with set to undefined

gideonthomas:
We can't just set to `undefined` because Bookshelf uses `hasOwnProperty` to check if a property exists or not.


cadecairos:
LGTM

Issue #237, opened 2017-02-02
Allow disabling the remix button


gideonthomas:
TODO:
- Add a column to the "projects" table that tracks whether the remix button is shown or not (default true)
- Accept a query string parameter for `/publish`. Based on the param, edit the "projects" table.

I think those are the broad strokes of what needs to be done.

@humphd do you think we should completely disable remixing (you can't remix even with the link) based on that toggle or just hide the UI (because that might change the steps above)?

Related: https://github.com/mozilla/thimble.mozilla.org/issues/1623

Pomax:
Don't we already have this in the form of the `readonly` column, used for goggles projects?

humphd:
Oh, it would be nice if we could re-use `readonly`.  @gideonthomas I think our goal is hiding that Remix stuff, I don't think it matters how we do it.

gideonthomas:
Nice catch @Pomax! Totally forgot about that! I think only the second step needs to be done in that case. And yeah, let's just hide the remix button instead of disabling it completely (since it's easier).

PR #238, opened 2017-02-04
Set readonly on projects during publish

gideonthomas:
Fix for #237 so that on publish, we can specify if the project is readonly or not.

### DIFF
--- a/api/classes/publisher.js
+++ b/api/classes/publisher.js
@@ -201,6 +201,18 @@ class BasePublisher {
     });
   }
 
+  updateProjectReadOnlyProperty(readonly) {
+    if (typeof readonly !== `boolean`) {
+      return Promise.resolve();
+    }
+
+    return projectsQueryBuilder
+    .updateOne(this.project.id, { readonly })
+    .then(() => {
+      this.project.readonly = readonly;
+    });
+  }
+
   setRemixDataForPublishedProject() {
     this.remixData = {
       projectId: this.publishedProject.id,
@@ -371,9 +383,10 @@ class Publisher extends BasePublisher {
     this.project = project.toJSON();
   }
 
-  publish() {
+  publish(readonly) {
     return Promise.resolve()
     .then(() => this.fetchUserForProject())
+    .then(() => this.updateProjectReadOnlyProperty(readonly))
     .then(() => this.createOrUpdatePublishedProject())
     .then(() => this.setRemixDataForPublishedProject())
     .then(() => this.uploadNewFiles())
--- a/test/lib/fixtures/projects/publish.js
+++ b/test/lib/fixtures/projects/publish.js
@@ -1,0 +1,31 @@
+var retrieveTestProjects = require('./test-projects');
+
+var published = {};
+var userToken = {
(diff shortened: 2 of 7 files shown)


api/classes/publisher.js
```diff
@@ -201,6 +201,18 @@ class BasePublisher {
     });
   }
 
+  updateProjectReadOnlyProperty(readonly) {
+    if (typeof readonly !== `boolean`) {
```
Pomax: we don't want to coerce off a possible string?

api/classes/publisher.js
```diff
@@ -201,6 +201,18 @@ class BasePublisher {
...
+      return Promise.resolve();
+    }
+
+    return projectsQueryBuilder
+    .updateOne(this.project.id, { readonly })
+    .then(() => {
+      this.project.readonly = readonly;
```
Pomax: would it not make more sense to refetch the project, rather than assume only the `.readonly` property got updated? This is async code, after all.

test/lib/fixtures/projects/publish.js
```diff
@@ -0,0 +1,31 @@
+var retrieveTestProjects = require('./test-projects');
+
+var published = {};
+var userToken = {
+  authorization: 'token ag-dubs'
```
Pomax: can we remove "specific people" from this codebase and instead use generic names or "testuser" or something? In part because people leave an org but more importantly because people new to the code will have no idea whether something is a username if it's a non-obvious sequence of letters that look like it might be meaningful code, such as with 'ag-dubs'

api/classes/publisher.js
gideonthomas: Want to try to avoid unnecessary calls to the db, also since `this.project` is not a model but the json representation of a model.

api/classes/publisher.js
gideonthomas: when the request comes in, because we have Joi validation, it automatically converts it to a boolean value so this would never be a string (it can only be `null` or `undefined`).

test/lib/fixtures/projects/publish.js
gideonthomas: I'll add another commit to fix this

gideonthomas: I take that back! I can file a follow-up bug for this but I tried implementing this and it requires a ton of code changes because we rely on the seed data we use for regular usage of publish.

gideonthomas:
@Pomax re-r?

test/lib/fixtures/projects/publish.js
Pomax: filed https://github.com/mozilla/publish.webmaker.org/issues/239 for you.

api/classes/publisher.js
Pomax: gotcha

test/lib/fixtures/projects/publish.js
gideonthomas: thank you

gideonthomas:
Thanks for the review @Pomax

Issue #240, opened 2017-02-11
Update html files during publish if project metadata changed after previous publish


gideonthomas:
Although https://github.com/mozilla/thimble.mozilla.org/issues/719 has been fixed, it doesn't actually update the published project itself, since on publish, we check from file changes only and only update the files on S3 that have changed since the last publish. We don't check for changes to project metadata and hence, we don't update the remix metadata that is stored in each html file on S3. This results in stale remix metadata being used in the published project.

We should diff the project and publishedProject metadata [here](https://github.com/mozilla/publish.webmaker.org/blob/master/api/classes/publisher.js#L245) to check if there were any changes, and if there were, in [this function](https://github.com/mozilla/publish.webmaker.org/blob/master/api/classes/publisher.js#L287-L330) we should force upload html files.

Pomax:
this might be an interesting bug for a CDOT student - comparing two objects is not super complicated, and a good thing to know how to do because you end up using it a lot in other projects

lkisac:
Hi @gideonthomas, I would like to start working on this bug

gideonthomas:
@lkisac sure thing (just a cautionary warning: this might be a slightly challenging bug)! Give me a little bit of time to figure out exactly what needs to be done here and I can add instructions below.

lkisac:
Hi @gideonthomas   I've been working on this bug.  I have an idea how to compare the objects, but I've been having a hard time debugging the lines of code mentioned above.  When I hit save and publish a project, I don't seem to get the breakpoint hit for the `createOrUpdatePublishedProject()` method on [this](https://github.com/mozilla/thimble.mozilla.org/blob/master/services/publish.webmaker.org/api/classes/publisher.js#L228) line, nor the [call](https://github.com/mozilla/thimble.mozilla.org/blob/master/services/publish.webmaker.org/api/classes/publisher.js#L390) to the method.  

I spent some time setting up my VS Code Chrome Debugger but had some issues with hitting breakpoints due to source mapping issue so I put that aside and am using Chrome DevTools but still breakpoints are not hit.  Since I don't have access to S3, can you recommend an approach to testing this?

lkisac:
Hi @gideonthomas  any updates?

gideonthomas:
[...]

PR #242, opened 2017-03-22
Fix 240 - Update html files during publish if project metadata change…

lkisac:
Fixes Issue #240

### DIFF
--- a/api/classes/publisher.js
+++ b/api/classes/publisher.js
@@ -290,7 +290,7 @@ class BasePublisher {
   }
 
   uploadModifiedFiles() {
-    const publishedProjectId = this.publishedProjectId;
+    const publishedProjectId = this.publishedProject.id;
     const remixData = this.remixData;
     const fileRoot = this.publishRoot;
     const username = this.user.name;
@@ -342,7 +342,7 @@ class BasePublisher {
 
         delete publishedFile.oldPath;
 
-        if (oldPath === publishedFile.path) {
+        if (!oldPath || oldPath === publishedFile.path) {
           return updateModelAndUpload(publishedFile);
         }
 
--- a/api/classes/publisher.js
+++ b/api/classes/publisher.js
@@ -314,7 +314,7 @@ class BasePublisher {
     }
 
     return publishedFilesQueryBuilder
-    .getAllModifiedFiles(this.publishedProject.id)
+    .getAllModifiedFiles(publishedProjectId)
     .then(publishedFiles => {
       if(!this.remixDataChanged) {
         return publishedFiles;
--- a/api/classes/publisher.js
+++ b/api/classes/publisher.js
@@ -322,7 +322,7 @@ class BasePublisher {
         return publishedFiles;
       }
 
-      var ignorePaths;
+      var ignorePaths = [];
 
       if(publishedFiles.length) {
         ignorePaths = publishedFiles.map(publishedFile => publishedFile.oldPath);
--- a/api/classes/publisher.js
+++ b/api/classes/publisher.js
@@ -20,8 +20,6 @@ const publishedFilesQueryBuilder = require(`../modules/publishedFiles/model`).pr
 
 const ROOT_URL = `/`;
 
-var updateHtmlFilesWithNewRemixData = false;
-
 /*
  * Utility functions
  */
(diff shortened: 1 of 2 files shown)


api/classes/publisher.js
```diff
@@ -310,6 +318,23 @@ class BasePublisher {
     return publishedFilesQueryBuilder
     .getAllModifiedFiles(this.publishedProject.id)
     .then(function(publishedFiles) {
+      if(!updateHtmlFilesWithNewRemixData) {
+        return publishedFiles;
+      }
+
+      var ignorePaths;
```
Pomax: given that this value gets passed into `getAllFilesMatchPath` a few lines below, what should its default value be?

flukeout:
Ping @gideonthomas or @cadecairos - either one of you want to review this?

api/classes/publisher.js
```diff
@@ -236,6 +238,11 @@ class BasePublisher {
     return this.fetchPublishedProject()
     .then(function(publishedProject) {
```
gideonthomas: change this to `.then(publishedProject => {`

api/classes/publisher.js
```diff
@@ -236,6 +238,11 @@ class BasePublisher {
...
       if (publishedProject) {
+        // Check if project title/description is different from publishedProject title/description
+        // If it is, set flag to indicate that html files of publishedProject need to be updated with new remix metadata
+        if (project.title !== publishedProject.title || project.description !== publishedProject.description) {
+          updateHtmlFilesWithNewRemixData = true;
```
gideonthomas: this should be `this.update...`

gideonthomas: and change the flag to `remixDataChanged` instead.

api/classes/publisher.js
```diff
@@ -310,6 +318,23 @@ class BasePublisher {
     return publishedFilesQueryBuilder
     .getAllModifiedFiles(this.publishedProject.id)
     .then(function(publishedFiles) {
```
gideonthomas: same thing here:
`.then(publishedFiles => {`

and inside you can use `this.update...`

api/classes/publisher.js
```diff
@@ -20,6 +20,8 @@ const publishedFilesQueryBuilder = require(`../modules/publishedFiles/model`).pr
 
 const ROOT_URL = `/`;
 
+var updateHtmlFilesWithNewRemixData = false;
```
gideonthomas: this needs to be removed.

api/classes/publisher.js
lkisac: Thanks for the feedback @gideonthomas.  I've made the changes and tested, it's working well on my end.  Please let me know if there are any issues

gideonthomas:
Hmm, looks like this fix didn't work.

I'm gonna debug to see why

api/classes/publisher.js
```diff
@@ -285,6 +290,7 @@ class BasePublisher {
   }
 
   uploadModifiedFiles() {
+    const publishedProjectId = this.publishedProjectId;
```
gideonthomas: this is incorrect! Look at how we access the publishedProject id below.

lkisac:
Thanks for the review @gideonthomas   I've made the necessary changes

gideonthomas:
@lkisac awesome thanks! Can you do one more small change? Can you make line 317 also use the variable you created for storing the published project id instead of accessing it directly?

lkisac:
Thanks @gideonthomas  I've changed line 317.  Before I commit, I noticed some of the other methods below `deletePublishedProject` ([deletePublishedFiles](https://github.com/mozilla/publish.webmaker.org/blob/master/api/classes/publisher.js#L348-L359), [deleteOldFiles](https://github.com/mozilla/publish.webmaker.org/blob/master/api/classes/publisher.js#L361-L378)) are using `this.publishedProject.id`.  Should I set a `const publishedProjectId` for those methods as well?

gideonthomas:
Let's keep this PR isolated to the changes for the issue it addresses so I don't think we should change those two methods.

Pomax:
Yeah, always keep work isolated as much as possible: if you see a thing that needs to be fixed "as well", file a new issue that explicitly mentions those specific things, and then file a separate PR that addresses only those changes. This keeps things cleanly separated and easy to track.

lkisac:
@Pomax  Absolutely.  It was something I noticed along side the last fix, while forgetting about the scope of the actual PR :/  I will file a new one for it, thanks!

lkisac:
Awesome, thanks for the merge @gideonthomas!

PR #246, opened 2017-05-30
Update Knex and Bookshelf

gideonthomas:
I updated `code` and `lab` as well to reflect the ORM updates. They should be inconsequential since they're just test framework updates.

`--leaks` needs to be passed into the tests now because of failures in Bookshelf's build (using babel). Tracking issue is https://github.com/tgriesser/bookshelf/issues/1191


gideonthomas:
Tested all workflows with Thimble and it works.

PR #248, opened 2017-06-01
Cache files for the remix case

gideonthomas:
This patch was becoming unwieldy so I decided to cache publishedFiles in the next patch. In this one, we only cache files when they are created on remixing a project so that we read from cache when we get all the files for the newly remixed project.

### DIFF
--- a/api/classes/prerequisites.js
+++ b/api/classes/prerequisites.js
@@ -85,7 +85,7 @@ class Prerequisites {
         const usernameParam = request.auth.credentials.username;
 
         const result = Promise.fromCallback(next => {
-          return request.server.methods.cache.user(usernameParam, next);
+          return request.server.methods.user(usernameParam, next);
         })
         .then(authenticatedUser => {
           if (!authenticatedUser) {
--- a/api/index.js
+++ b/api/index.js
@@ -5,7 +5,7 @@ const Hoek = require(`hoek`);
 exports.register = function api(server, options, next) {
   const cacheEnabled = !!(process.env.REDIS_URL);
 
-  function api() {
+  function routeSetup() {
     server.register([{
       register: require(`./modules/users/routes`)
     }, {
@@ -79,14 +79,14 @@ exports.register = function api(server, options, next) {
           }
         }
 
-        server.method(cache.name, () => cache.run(), cacheConfig);
+        server.method(cache.name, cache.run.bind(cache), cacheConfig);
       });
     });
   }
 
   if (!cacheEnabled) {
     cacheSetup();
-    api();
+    routeSetup();
     return;
   }
 
@@ -107,7 +107,7 @@ exports.register = function api(server, options, next) {
       Hoek.assert(catboxEngine.client && typeof catboxEngine.client === `object`, `Can't find redis client in Catbox engine`);
       clearInterval(intervalId);
       cacheSetup(catboxEngine.client);
-      api();
+      routeSetup();
     }
   }, 500);
 };
--- a/api/index.js
+++ b/api/index.js
@@ -15,7 +15,7 @@ exports.register = function api(server, options, next) {
         };
       }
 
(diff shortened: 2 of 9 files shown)


api/classes/prerequisites.js
```diff
@@ -85,7 +85,7 @@ class Prerequisites {
         const usernameParam = request.auth.credentials.username;
 
         const result = Promise.fromCallback(next => {
-          return request.server.methods.cache.user(usernameParam, next);
```
humphd: Why did you stop using cache here?

api/modules/files/cache.js
```diff
@@ -0,0 +1,48 @@
...
+
+    return Files.query({
+      where: {
+        id: fileId
+      },
+      columns: [`buffer`]
+    })
+    .fetch()
+    .then(file => {
+      console.log(`Typeof buffer: `, typeof file.get(`buffer`));
```
humphd: Should remove this

api/classes/prerequisites.js
gideonthomas: It was just the method name, it didn't actually represent cache. I removed it because otherwise the method to drop a key would be `request.server.methods.cache.user.cache.drop`. It seemed redundant so I remover the first `cache`. It still caches the returned value of the method since we have a cache config setup for the method

api/modules/files/cache.js
gideonthomas: will do once the patch is done. keeping it for now since this is a wip pr.

humphd:
why this PR is going to be so useful:

screen shot 2017-06-03 at 11 54 20 am

gideonthomas:
This is ready for review (and testing, I added logs to help - will remove them before merging). I updated my PR description to scope down this fix.

api/modules/files/cache.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    return this.cache.getBuffer(`file:${fileId}`)
+    .then(cachedFileBuffer => {
+      if (!cachedFileBuffer) {
+        console.log(`DB HIT FOR FILE`);
+        return getBuffer(fileId, next);
+      }
+
+      console.log(`CACHE HIT for file`);
```
gideonthomas: this is only for testing won't be in the commits merged in

api/modules/files/cache.js
```diff
@@ -4,40 +4,73 @@ const Files = require(`./model`);
...
+     * Putting the file buffer into cache
+     */
+
     if(typeof fileBuffer !== `function`) {
-      return next(null, fileBuffer);
+      if(!this.cache) {
+        return next(null, fileBuffer);
+      }
+
+      return this.cache.setex(`file:${fileId}`, DEFAULT_BUFFER_CACHE_EXPIRY_SEC, fileBuffer)
```
humphd: What about needing to inform redis underneath that we're using `Buffer` types as values?  Does it matter, or is it happening for free in the redis client catbox is using?

gideonthomas: ioredis apparently does this automatically. You only need to specify that you want a buffer when you `get` it from ioredis

api/index.js
```diff
@@ -1,42 +1,53 @@
...
+          cache: cache.config
         };
+
+        if (cache.generateKey) {
+          cacheConfig.generateKey = cache.generateKey;
+        }
       }
 
-      server.method(`cache.${Cache.name}`, Cache.run, cacheConfig);
+      server.method(cache.name, cache.run.bind(cache), cacheConfig);
```
gideonthomas: it is a prototype function of cache that relies on `this`. Since the bind is a one time cost only (when the server starts), there isn't much of a perf hit. Ideally with ES6 this would just be `(...args) => cache.run(...args)` but node 4 won't let us do that.

humphd:
Love this.

PR #249, opened 2017-06-04
Cache published files during remix

gideonthomas:
Builds on https://github.com/mozilla/publish.webmaker.org/pull/248

Had to do a bunch of extra work here since the way we were getting the published files from the db during remix was via a single db query. With caching now, since the publishedFiles is a <b>collection</b> and <b>contains mixed data (json + buffers)</b>, I had to write some extra logic to cache them separately.

### Code logic
Check if the publishedFiles metadata (all metadata stored in one key identified by the published project id) is in cache
   - YES - Fetch it and loop through each publishedFile's metadata and check if the corresponding buffer exists in cache (key being the publishedFileId)
      - YES - [FULL Cache hit / NO DB hit] Fetch it and augment the metadata object with this buffer and add it to the array that will contain all the publishedFiles.
      - NO - [PARTIAL Cache hit / DB hit per missing buffer from cache] Fetch the buffer from the database, write it to cache, and augment the metadata object with this buffer and add it to the array that will contain all the publishedFiles.
      - FINALLY - Return the publishedFiles array to the caller.
   - NO - [NO Cache hit / FULL DB hit - single query to get all publishedFiles] Fetch the publishedFiles (including buffers) from the database. Loop through each model and for each model, 1. Add to an array that will contain only the publishedFile metadata, 2. Write the buffer to the cache, 3. Convert the model to a POJO and add it to an array that will contain the publishedFiles that will be returned to the caller. Once the loop is finished, write the constructed array that contains only the publishedFile metadata to cache and then return the publishedFiles to the caller.

I'm still working on simplifying the code to reflect this logic so this is still a WIP.

### DIFF
--- a/api/modules/files/cache.js
+++ b/api/modules/files/cache.js
@@ -45,12 +45,11 @@ class RemixedFileCreationCache extends BaseCache {
 
     if (typeof fileBuffer !== `function`) {
       if (!this.cache) {
-        console.log(`NO cache, buffer not set in cache`);
         return next(null, fileBuffer);
       }
 
       return this.cache.setex(`file:${fileId}`, DEFAULT_BUFFER_CACHE_EXPIRY_SEC, fileBuffer)
-      .then(() => { console.log(`Setting cache`); next(null, fileBuffer); })
+      .then(() => next(null, fileBuffer))
       .catch(next);
     }
 
@@ -62,18 +61,14 @@ class RemixedFileCreationCache extends BaseCache {
      */
 
     if (!this.cache) {
-      console.log(`NO Cache. Getting buffer from DB.`);
       return getBuffer(fileId, next);
     }
 
     return this.cache.getBuffer(`file:${fileId}`)
     .then(cachedFileBuffer => {
       if (!cachedFileBuffer) {
-        console.log(`DB HIT FOR FILE`);
         return getBuffer(fileId, next);
       }
-
-      console.log(`CACHE HIT for file`);
 
       next(null, cachedFileBuffer);
     })
--- a/api/modules/publishedFiles/cache.js
+++ b/api/modules/publishedFiles/cache.js
@@ -6,7 +6,8 @@ const PublishedFiles = require(`./model`);
 
 const BaseCache = require(`../../classes/base_cache`);
 
-const PUBLISHED_FILE_CACHE_EXPIRY_MS = 10 * 60 * 1000;
+const PUBLISHED_FILE_CACHE_EXPIRY_SEC = 10 * 60;
+const META_KEY_PREFIX = `publishedProjectFilesMeta`;
 
 class PublishedFilesByPublishedProjectCache extends BaseCache {
   constructor(server) {
@@ -35,13 +36,13 @@ class PublishedFilesByPublishedProjectCache extends BaseCache {
       }
     })
     .fetchAll()
-    .then(publishedFileModels => {
+    .then(result => {
       // For each published file from the db:
       // 1. Accumulate a simple js object copy of it in an array to be
       //    returned back,
       // 2. Accumulate the metadata in an array, and
       // 3. Cache its buffer
-      return Promise.map(publishedFileModels, publishedFileModel => {
+      return Promise.map(result.models, publishedFileModel => {
         const publishedFile = publishedFileModel.toJSON();
 
         publishedFiles.push(publishedFile);
@@ -52,7 +53,7 @@ class PublishedFilesByPublishedProjectCache extends BaseCache {
 
         return this.cache.setex(
           `${bufferKeyPrefix}:${publishedFile.id}`,
-          PUBLISHED_FILE_CACHE_EXPIRY_MS,
+          PUBLISHED_FILE_CACHE_EXPIRY_SEC,
           publishedFile.buffer
         );
       });
@@ -61,8 +62,8 @@ class PublishedFilesByPublishedProjectCache extends BaseCache {
       // Cache the accumulated metadata
       return this.cache.setex(
         metaKey,
-        PUBLISHED_FILE_CACHE_EXPIRY_MS,
-        publishedFilesMeta
+        PUBLISHED_FILE_CACHE_EXPIRY_SEC,
+        JSON.stringify(publishedFilesMeta)
       );
     })
     // Return the array of simple js objects of published files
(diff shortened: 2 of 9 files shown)


api/modules/files/cache.js
```diff
@@ -45,12 +45,11 @@ class RemixedFileCreationCache extends BaseCache {
 
     if (typeof fileBuffer !== `function`) {
       if (!this.cache) {
-        console.log(`NO cache, buffer not set in cache`);
         return next(null, fileBuffer);
       }
 
       return this.cache.setex(`file:${fileId}`, DEFAULT_BUFFER_CACHE_EXPIRY_SEC, fileBuffer)
```
humphd: Sometimes you set your TTL as `SEC` and other times `MS`.  Is it different for these various APIs?

api/modules/publishedFiles/cache.js
```diff
@@ -0,0 +1,163 @@
...
+        PUBLISHED_FILE_CACHE_EXPIRY_MS,
+        publishedFileBuffer
+      )
+      .then(() => publishedFileBuffer);
+    });
+  }
+
+  // PARTIAL or FULL Cache hit / PARTIAL or NO DB hit
+  _getAllBuffersFromCache(bufferKeyPrefix, publishedFilesMeta) {
+    // Accumulate all the POJO representations of the published files in an
```
humphd: What does `POJO` stand for?

api/modules/publishedFiles/cache.js
```diff
@@ -0,0 +1,163 @@
...
+      const publishedFileId = publishedFileMeta.id;
+
+      // Attempt to get the buffer from the cache
+      return this.cache.getBuffer(`${bufferKeyPrefix}:${publishedFileId}`)
+      .then(publishedFileBuffer => {
+        if (publishedFileBuffer) {
+          // FULL Cache hit / NO DB hit
+          // Combine the metadata and buffer and return a POJO representation
+          // of it
+          console.log(`Cache hit for buffer`);
```
humphd: Remove

api/modules/publishedFiles/cache.js
```diff
@@ -0,0 +1,163 @@
...
+          publishedFiles.push(publishedFileMeta);
+
+          return publishedFiles;
+        }
+
+        // Partial Cache/DB hit
+        // Get the buffer from the db
+        return this._getBufferFromDB(bufferKeyPrefix, publishedFileId)
+        .then(publishedFileBufferFromDB => {
+          console.log(`DB hit for buffer`);
```
humphd: Remove

api/modules/publishedFiles/cache.js
```diff
@@ -0,0 +1,163 @@
...
+
+          return publishedFiles;
+        });
+      });
+    }, []);
+  }
+
+  run(publishedProjectId, next) {
+    if (!this.cache) {
+      console.log(`Cache not enabled`);
```
humphd: Remove

api/modules/publishedFiles/cache.js
```diff
@@ -0,0 +1,163 @@
...
+    const metaKey = `publishedProjectFilesMeta:${publishedProjectId}`;
+    const bufferKeyPrefix = `publishedFile`;
+
+    return this.cache.get(metaKey)
+    .then(publishedFilesMeta => {
+      if (!publishedFilesMeta) {
+        return this._getAllDataFromDB(metaKey, bufferKeyPrefix, publishedProjectId);
+      }
+
+      console.log(`Cache hit for metadata`);
```
humphd: Remove

api/modules/publishedFiles/cache.js
```diff
@@ -0,0 +1,163 @@
...
+        return this._getBufferFromDB(bufferKeyPrefix, publishedFileId)
+        .then(publishedFileBufferFromDB => {
+          console.log(`DB hit for buffer`);
+          publishedFileMeta.buffer = publishedFileBufferFromDB;
+          publishedFiles.push(publishedFileMeta);
+
+          return publishedFiles;
+        });
+      });
+    }, []);
```
humphd: Can you add a comment to explain this `[]` at the end?

api/modules/files/cache.js
gideonthomas: yeah! catbox takes it as ms whereas ioredis takes it as secs

api/modules/publishedFiles/cache.js
gideonthomas: plain old javascript object :D

api/modules/publishedFiles/cache.js
```diff
@@ -0,0 +1,167 @@
...
+      console.log(`Cache hit for metadata`);
+      return this._getAllBuffersFromCache(bufferKeyPrefix, publishedFilesMeta);
+    })
+    .then(publishedFiles => next(null, publishedFiles))
+    .catch(next);
+  }
+}
+
+module.exports = {
+  PublishedFilesByPublishedProjectCache
```
cadecairos: `this.server.methods.publishedFilesByPublishedProject is not a function`

api/modules/publishedProjects/controller.js
```diff
@@ -45,21 +44,18 @@ class Remix {
...
 
   _getFilesToRemix() {
-    return PublishedFilesModel.query({
-      where: {
-        published_id: this.publishedProjectsModel.get(`id`)
-      }
-    })
-    .fetchAll();
+    return Promise.fromCallback(next => {
+      return this.server.methods.publishedFilesByPublishedProject(this.publishedProjectsModel.get(`id`), next);
```
cadecairos: or rather: `this.server.methods.publishedFilesByPublishedProject is not a function`

api/modules/publishedFiles/cache.js
```diff
@@ -0,0 +1,167 @@
...
+      }
+    })
+    .fetchAll()
+    .then(publishedFileModels => {
+      // For each published file from the db:
+      // 1. Accumulate a simple js object copy of it in an array to be
+      //    returned back,
+      // 2. Accumulate the metadata in an array, and
+      // 3. Cache its buffer
+      return Promise.map(publishedFileModels, publishedFileModel => {
```
cadecairos: New error:

`TypeError: Promise.map is not a function`

api/modules/publishedFiles/cache.js
```diff
@@ -0,0 +1,169 @@
+"use strict";
+
+const Promise = require(`bluebird`);
```
Pomax: Is this because of Node 4?

gideonthomas: not exactly. bluebird has some pretty useful promise based functions that help us handle collections. Once we switch to node 6, I would use generators instead.

gideonthomas:
@cadecairos sorry, this should be ready to try again.

gideonthomas:
ugh travis

PR #253, opened 2017-06-12
Add caching for getting a single file

gideonthomas:
Note that this only works for the remix case and is meant to be run with https://github.com/mozilla/thimble.mozilla.org/pull/2262.

So, when you remix a project in that patch, you should see `Fetched from cache` logs but otherwise, you should see `Fetched from DB` logs (which I will remove before we merge)

### DIFF
--- a/api/modules/files/controller.js
+++ b/api/modules/files/controller.js
@@ -44,7 +44,7 @@ class FilesController extends BaseController {
   getAllAsTar(request, reply) {
     const files = request.pre.records.models;
     const tarStream = Tar.pack();
-    const fileCache = request.server.methods.createdRemixFile;
+    const fileCache = request.server.methods.file;
 
     function processFile(file) {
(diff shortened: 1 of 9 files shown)


humphd:
I'm having a really hard time testing this locally.  @cadecairos, I think I'm going to have to punt the review totally to you.

api/modules/files/controller.js
```diff
@@ -73,12 +73,26 @@ class FilesController extends BaseController {
...
+  }
+
   create(request, reply) {
     return super.create(request, reply, formatResponse);
   }
 
   update(request, reply) {
-    return super.update(request, reply, formatResponse);
+    return Promise.fromCallback(next => {
+      return request.server.methods.file.cache.drop(request.params.id, next);
```
cadecairos: While it's probably no big deal, we can safely drop the cache and not block the response to the client. Just a recommendation.

PR #255, opened 2017-06-16
Fix mozilla/thimble.mozilla.org#2295 - Update file on create route if path already exists

gideonthomas:
See mozilla/thimble.mozilla.org#2295


api/modules/files/controller.js
```diff
@@ -3,6 +3,7 @@
 const Promise = require(`bluebird`);
 const fs = require(`fs`);
 const Tar = require(`tar-stream`);
+const Boom = require(`boom`);
```
cadecairos: 💥

PR #260, opened 2017-07-26
Fix #2288: Fixed the project description

atiqueahmedziad:
Fix mozilla/thimble.mozilla.org/issues/2288


atiqueahmedziad:
@gideonthomas can you please check if it is alright?

PR #268, opened 2017-08-23
Fix promise resolution in route handlers

gideonthomas:
Unfortunately, I don't have a way to test this locally. This only breaks on prod and staging.

Before, we did not return a promise chain from some of our handlers and called the reply interface in an inherited handler which I think made hapi not wait on the initial promise to complete. This makes sure that we always only pass in one promise chain into the reply interface and also separates the create/update/delete db logic from the actual HTTP logic.

This should hopefully fix mozilla/thimble.mozilla.org#2441

### DIFF
--- a/api/classes/base_controller.js
+++ b/api/classes/base_controller.js
@@ -18,8 +18,7 @@ class BaseController {
     this.Model = Model;
   }
 
-  /* eslint no-unused-vars: ["error", { "argsIgnorePattern": "^request$" }]*/
-  formatRequestData(request) {
+  formatRequestData(request) { // eslint-disable-line no-unused-vars
     // Abstract base method; formats data for database entry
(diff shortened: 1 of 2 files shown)


flukeout:
I'll check if the problem is happening on staging to. If it is we can maybe deploy it there for testing to make sure this is helping.

flukeout:
@gthomas don't merge until I get a chance to check on staging (later tonight) - I'll comment here once I do so. Cheers.

flukeout:
@gideonthomas I can't reproduce the error in staging. Can you?

flukeout:
Anyway - I guess let's merge it and test in production tomorrow?

gideonthomas:
@flukeout yeah, my guess is that since we don't have any load on staging, you won't notice it there either. I think with heavier load, requests take longer to complete and having promise chain errors is more noticable.

PR #270, opened 2018-09-13
Add routes and middleware to begin exporting projects

gideonthomas:
Fix mozilla/thimble.mozilla.org#2668

This PR does 3 things:
1. `/projects/<project id>/export/start/` route that generates a token allowing access to export a specific project
2. `/publishedprojects/<published project id>/export/start/` route that generates a token allowing access to export a specific published project
3. Authorization middleware to check if the project (or published project) id in the URL and the token provided matches what’s in the data store (this will be used for the other export routes).

To test:
1. Run id/login/thimble (no need to run brackets). Add ``` console.log(`Token: ${token}`);``` to [this line in the Thimble code](https://github.com/mozilla/thimble.mozilla.org/blob/master/server/lib/middleware.js#L99).
2. Run a redis server and set the `REDIS_URL=http://0.0.0.0:6379` in your env for the publish server.
3. Run publish and login to Thimble. Copy the token that was logged in the thimble server. Make sure that project with id 1 is shown in your "My Projects" list in Thimble.
4. Run the following command and replace `<token>` with the copied token. Make sure the response you get back has a token in it (and copy the token value).
```
curl -X POST -H 'Authorization: token <token>' -v -i 'http://localhost:2015/projects/1/export/start'
```
5. **BONUS**: Check if the token was added to redis:
   - In a terminal window opened in your publish server folder, run `node`. Then type in the following code replacing `<token>` with the token you copied in the previous step: <br> `require('crypto').createHmac('sha256', 'randomSaltString').update(<token>).digest('hex')`
   - Copy the result and run `redis-cli`
   - In the redis terminal, run `get <saltedToken>` and replace `<saltedToken>` with the value you copied in the previous step. Make sure the returned value is `1`.


api/modules/projects/controller.js
```diff
@@ -185,6 +185,18 @@ class ProjectsController extends BaseController {
...
   }
+
+  exportStart(request, reply) {
+    const project = request.pre.records.models[0];
+    const exportProjectCache = request.server.methods.exportProject;
+    const projectIdentifier = { id: project.get(`id`) };
+
+    return reply(
+      Promise.fromCallback(next => exportProjectCache(projectIdentifier, next))
+      .then(token => request.generateResponse({ token }).code(200))
```
cadecairos: nit: 201 response code seems more appropriate

gideonthomas:
Perfect! Thanks for the review @cadecairos!

